PR #62, opened 2026-03-14
fix: doubly-nested correlated scalar subquery + trigger fixes + cross-join fix

carli2:
## Summary

- **fix: doubly-nested correlated scalar subquery** — `SELECT (SELECT f.filename FROM nc_files f WHERE f.ID = (SELECT r.file FROM nc_revision r WHERE r.doc = nc_doc.ID ...) ...) FROM nc_doc` was returning NULL instead of the correct value. Root cause: `untangle_query`'s `outer_schemas` was spread via `append` across individual list elements instead of passed as a single 10th argument. Fix: call `untangle_query` directly with explicit `(nth subquery N)` args so `outer_schemas` is passed as a single assoc list value. Added `_unnesting_enabled` guard to disable Neumann unnesting inside subquery contexts (prevents multi-group-stage errors). Test 102 covers all three doubly-nested patterns.
- **fix: CREATE TRIGGER panics when body references table not yet created** — triggers now compile gracefully when referenced tables don't exist yet (test 101)
- **fix: cross-join correlated IN subquery** — correlated IN subqueries inside cross-joins now return correct results (test 100)
- **fix: CREATE TRIGGER drops visible=true arg** — fixed argument handling in trigger DDL
- **feat: INSERT ... VALUES SELECT** + TIMESTAMPDIFF function support
- Various AUTO_INCREMENT and trigger persistence fixes (from earlier commits merged into branch)

## Test plan

- [ ] `python3 run_sql_tests.py tests/102_nested_correlated_subquery.yaml PORT --connect-only` → 3/3
- [ ] `python3 run_sql_tests.py tests/100_crossjoin_correlated_in.yaml PORT --connect-only` → passes
- [ ] `python3 run_sql_tests.py tests/101_trigger_body_compile_error.yaml PORT --connect-only` → passes
- [ ] Full pre-commit test suite (`git commit`) — CI green


carli2:
Superseded by unnesting-arbitrary-queries which implements the full Neumann dependent-join flattening. The only unique commit (f64e6b7, the apply+append calling convention fix) has been communicated as a direct fix to that branch.

PR #192, opened 2026-04-12
fix: re-enable scan_batch rewrite after catalog merge (+67% equi-join)

carli2:
## Summary

- The catalog merge (096b897ce) silently disabled the nested-loop → batch-nested-loop rewrite for `scan`; every equi-join fell back to plain nested loop.
- Root cause: `tryScanBatchRewrite` guards on `v[2].IsCustom(TagTable)`, but at optimizer time `v[2]` is a symbol `tbl:schema:name` (pre-resolved reference) or a `(table schema tbl)` codegen list — never the runtime TagTable value. Guard never matched → rewrite never fired.
- Secondary latent bug in `scan_batch`'s runtime `Fn` handler: reduce/neutral/reduce2/isOuter were read from `layout.*Idx+1`, but the stride+batchdata insertion shifts those args by 2 positions, not 1. Invisible so far because the rewrite never produced scan_batch calls.

## Measurement

A/B with median-of-5 at 3 250 rows (pre-catalog commit `e4702619a` vs post-catalog `7e6a334ad` vs this PR):

| Test       | Baseline | Catalog (unfixed) | Catalog + fix | Δ vs baseline |
|------------|---------:|------------------:|--------------:|--------------:|
| COUNT      |  1.6 ms  |  1.5 ms           |  1.4 ms       |  -13% |
| SUM        |  1.3 ms  |  1.3 ms           |  1.3 ms       |   0%  |
| ORDER+LIMIT|  4.7 ms  |  5.4 ms           |  4.3 ms       |  -9%  |
| SCAN+FILTER|  1.3 ms  |  1.3 ms           |  1.3 ms       |   0%  |
| GROUP BY   | 46.6 ms  | 47.9 ms           | 47.7 ms       |  +2%  |
| **EQUI JOIN**| **22.0 ms** | **36.7 ms** ⚠️ | **22.0 ms** ✅ | 0% |
| RANGE JOIN | 24.6 ms  | 16.1 ms           | 23.4 ms       |  -5%  |
| **DELTA JOIN**| **26.8 ms** | **40.6 ms** ⚠️ | **26.7 ms** ✅ | 0% |

## Changes

- `storage/scan_batch_rewrite.go`: drop the `TagTable` guard on both outer and inner scan. v[2] is always a table reference of some shape; pass it through and let runtime resolve. Keep `tryScanOrderBatchRewrite` disabled pending separate review (has same latent index-drift + different ordered-result semantics).
- `storage/storage.go`: fix scan_batch Fn handler offsets — `+1` → `+2` for reduce/neutral/reduce2/isOuter, introduced a `sbShift` constant so the intent is visible.

## Test plan

- [x] `tests/90_perf_basic.yaml` EQUI/RANGE/DELTA JOIN pass post-fix with times matching pre-catalog
- [x] Full pre-commit test suite (all `[0-9][0-9]_*.yaml`) passes locally
- [ ] CI green

### DIFF
--- a/lib/dashboard.scm
+++ b/lib/dashboard.scm
@@ -366,7 +366,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 /* Metrics tracing: periodically insert rows into system_statistic.perf_metrics */
 (if (not (has? (show "system_statistic") "perf_metrics")) (begin
 	(print "creating table system_statistic.perf_metrics")
-	(eval (parse_sql "system_statistic" "CREATE TABLE perf_metrics(time text, cpu float, mem_available bigint, mem_total bigint, shard_memory bigint, shard_budget bigint, connections int, max_connections int, rps float, eps float) ENGINE=SLOPPY" (lambda (schema table write) true)))
+	(eval (parse_sql "system_statistic" "CREATE TABLE perf_metrics(time text, cpu float, mem_available bigint, mem_total bigint, shard_memory bigint, shard_budget bigint, connections int, max_connections int, rps float, eps float) ENGINE=SLOPPY" (lambda (schema tblname write) true)))
 ))
 
 /* self-scheduling tracing loop via setTimeout */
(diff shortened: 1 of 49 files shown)


carli2:
Superseded by #194 — this branch had 22 pre-merge commits from the catalog work that are already in master via #191; rebasing onto master produces the same diff from a clean branch.

PR #242, opened 2026-04-18
Optimize repeated queryplan term analysis

carli2:
## What
- replace deep `equal?` dedup in hot scalar/domain helper paths with serialized string keys
- pre-group aggregate scalar parts by alias instead of rescanning the same join-condition list per alias
- memoize top-level pure AST analysis helpers for repeated planner traversals:
  - `extract_tblvars`
  - `extract_columns_for_tblvar`
  - `extract_outer_columns_for_tblvar`
  - `unnest_expr_outer_refs`

## Why
Large app queries compile through repeated planner traversals over the same nested scalar and permission terms. This PR reduces repeated term analysis without changing lowering semantics.

## Validation
- `python3 run_sql_tests.py tests/32_expr_subselects.yaml`
- `python3 run_sql_tests.py tests/117_dashboard_unnesting_regressions.yaml`
- `make test` was started locally and ran through the broad suite without failures in the observed output; CI will provide the definitive full run

## Cold benchmark
Fresh restart / fresh in-memory queryplan cache, same copied app data and badge query:
- `master`: `28.78s`
- this branch: `29.09s`

So the branch is currently correctness-safe but benchmark-neutral on that exact homepage badge query. I am still opening it because the optimization direction is general and the changes are isolated enough for CI review.

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -554,6 +554,29 @@ keep it in a local define instead of rebuilding it from scratch. */
 (define serialize_canonical_expr (lambda (expr)
 	(serialize (require_canonical_logical_expr "serialize_canonical_expr" expr))
 ))
+(define canonical_logical_expr_key (lambda (expr)
+	(serialize_canonical_expr (normalize_canonical_aliases expr))
+))
+(define keyed_dedup (lambda (items key_fn) (begin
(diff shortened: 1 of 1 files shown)


carli2:
nonsense

PR #271, opened 2026-07-24
Replace GLS with lock-free context frames

carli2:
## Summary
- vendor `github.com/jtolds/gls` under `third_party/gls` and replace it from `go.mod`
- replace upstream GLS mutexed mutable per-goroutine maps with CAS-published manager/slot tables and per-goroutine immutable context frames
- keep `gls.Go` parallelism: child goroutines inherit the same immutable frame pointer instead of copying or mutating shared `Values` maps
- add an anonymized SQL regression proving session variables survive parallel multishard scans

## Why
The upstream GLS implementation stores `map[uint]Values` behind a mutex, while `gls.Go` passes the parent `Values` map into `SetValues`. Parallel workers can then iterate and mutate the same map concurrently, which produced crashes like `fatal error: concurrent map iteration and map write`. The previous storage-level workaround avoided that by snapshotting context values at call sites, but it added the wrong abstraction in storage/compute code. This PR fixes the context structure itself.

## Concurrency/performance notes
- Read path: `GetValue` does stack-tag gid lookup, one atomic load of the immutable slot table, one map lookup, then frame lookup. No locks.
- Context entry: `SetValues` pushes a frame in the current goroutine and restores the previous pointer with `defer`; no inherited value-map copy.
- Fan-out: `gls.Go` captures and reuses the immutable frame pointer for each manager; no copied `Values` maps and no mutation of parent frames.
- The slot table uses copy-on-write CAS publication like MemCP's `NonLockingReadMap` pattern. Per-gid frame pointers are plain owner-goroutine pointers; active gids are unique and released through the existing id pool.
- `scm/session` was reviewed: it still uses its own `RWMutex` for session variable map access, but this PR removes the GLS-level global mutex/copy hazard without changing session semantics.

## Benchmarks
Compared against `origin/master` (`393f81645`) on AMD Ryzen 9 7900X3D, `go test ./storage -run '^$' -bench 'BenchmarkScanFixedCosts_WithGLS|BenchmarkScanFixedCosts_WithAutocommit' -benchtime=1s -count=5 -benchmem`:

- master `WithGLS`: ~20.3-20.9 us/op, 7910 B/op, 56 allocs/op
- branch `WithGLS`: ~20.8-21.7 us/op, 7614 B/op, 53 allocs/op
- master `WithAutocommit`: ~32.7-33.4 us/op, 9078 B/op, 72 allocs/op
- branch `WithAutocommit`: ~31.1-32.1 us/op, 8654 B/op, 68 allocs/op

The synthetic raw `SetValues` loop is still a few percent slower in ns/op, but allocation count and bytes are lower. The more realistic autocommit scan path is faster and allocates less.

## Validation
- `go test ./...` in `third_party/gls`
- `go test -race ./...` in `third_party/gls`
- `make`
- `python3 run_sql_tests.py tests/56_multishard.yaml`
- `go test ./storage -run 'TestGlobalAggregateComputeAndInsertDoNotDeadlock|TestFilteredComputeColumnConservativelyRecomputesRepeatedFilter'`
- `python3 run_sql_tests.py tests/84_incremental_invalidation.yaml`
- `python3 run_sql_tests.py tests/107_orc_trigger_dependencies.yaml`
- isolated reruns of `tests/64_order_by_complex.yaml` and `tests/65_trigger_subselect_scope.yaml` after the full shared run timed out

`make test` was also started. It progressed through many suites, then the shared server produced `No response` failures around `tests/64_order_by_complex.yaml` / `tests/65_trigger_subselect_scope.yaml`. Both suites passed immediately when rerun on fresh isolated servers. The SIGQUIT dump showed the server piled up timed-out POSTs after a long query, not a GLS lock wait.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -58,3 +58,5 @@ replace github.com/launix-de/NonLockingReadMap => ./third_party/NonLockingReadMa
 replace github.com/launix-de/go-mysqlstack => ./third_party/go-mysqlstack
 
 replace golang.org/x/text => github.com/carli2/text v0.34.1-0.20260305004517-20c7a406302a
+
+replace github.com/jtolds/gls => ./third_party/gls
--- a/tests/56_multishard.yaml
+++ b/tests/56_multishard.yaml
@@ -67,6 +67,19 @@ test_cases:
       data:
         - cnt: 999
 
+  - name: "Set session variable for parallel shard scan"
+    session_id: "ms_parallel_session"
+    sql: "SET @ms_min = 60000"
+    expect: {}
+
+  - name: "Session variable is visible in parallel shard scan"
+    session_id: "ms_parallel_session"
+    sql: "SELECT COUNT(*) AS cnt FROM ms_data WHERE val > @ms_min"
+    expect:
+      rows: 1
+      data:
+        - cnt: 999
+
   - name: "COUNT with BETWEEN filter across shards"
     sql: "SELECT COUNT(*) AS cnt FROM ms_data WHERE val BETWEEN 30000 AND 30009"
     expect:
--- a/third_party/gls/LICENSE
+++ b/third_party/gls/LICENSE
@@ -1,0 +1,18 @@
+Copyright (c) 2013, Space Monkey, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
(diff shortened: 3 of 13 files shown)


carli2:
Relevant end-to-end shard-parallel query benchmark, measured after the GLS rewrite because the microbenchmarks do not answer whether shard parallelism is preserved.

Setup:
- rows: 600,000 in `perf_data`, rebuilt/repartitioned after load
- branch commit: `57ebf3d33`
- master baseline: `393f81645`
- queries measured through HTTP `/sql`, 7 warm repetitions after one warmup
- `branch-1cpu` is enforced with `taskset -c 0`; `GOMAXPROCS=1` is not reliable here because the code/deps reset/use `runtime.NumCPU()`

Results:

| run | query | median wall | cpu/wall |
| --- | --- | ---: | ---: |
| branch all CPUs | `ORDER BY val DESC LIMIT 100` | 0.0724s | 11.74x |
| branch 1 CPU | `ORDER BY val DESC LIMIT 100` | 0.5367s | 1.01x |
| master all CPUs | `ORDER BY val DESC LIMIT 100` | 0.0736s | 11.55x |
| branch all CPUs | `WHERE val >= @zero ORDER BY val DESC LIMIT 100` | 0.1196s | 11.37x |
| branch 1 CPU | `WHERE val >= @zero ORDER BY val DESC LIMIT 100` | 0.7876s | 0.99x |
| master all CPUs | `WHERE val >= @zero ORDER BY val DESC LIMIT 100` | 0.1195s | 11.29x |
| branch all CPUs | `WHERE name LIKE @allpat ORDER BY val DESC LIMIT 100` | 0.0583s | 11.14x |
| branch 1 CPU | `WHERE name LIKE @allpat ORDER BY val DESC LIMIT 100` | 0.3112s | 1.00x |
| master all CPUs | `WHERE name LIKE @allpat ORDER BY val DESC LIMIT 100` | 0.0573s | 11.00x |

Interpretation:
- The branch still uses real shard parallelism: all-CPU runs consume about 11x CPU per wall second, while the same branch pinned to one CPU is about 4.5x-6.6x slower depending on query.
- The session-variable query is the relevant GLS case. It keeps the same parallel shape (`11.37x` cpu/wall) and has the same wall time as master (`0.1196s` branch vs `0.1195s` master).
- The GLS rewrite therefore does not visibly throttle shard parallel query execution in these end-to-end query measurements.

PR #272, opened 2026-07-25
Keep empty SELECT results distinct

carli2:
## Summary
- keep empty row-returning statements from being serialized as affected_rows=0
- track resultrow calls through a session-backed flag so callbacks inside lambdas update shared state
- add regressions for empty SELECT-style responses not containing affected_rows

## Tests
- python3 tools/lint_scm.py
- make
- python3 run_sql_tests.py tests/24_mysql_basic_compat.yaml
- ./git-pre-commit tests/24_mysql_basic_compat.yaml

Note: a later full pre-commit attempt was interrupted because I accidentally started full hooks in two worktrees at once; both used the fixed test port/data directory and interfered with each other. The targeted hook above completed successfully before the commit.

### DIFF
--- a/lib/sql.scm
+++ b/lib/sql.scm
@@ -20,6 +20,26 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 (import "sql-builtins.scm")
 (import "queryplan.scm")
 
+(define sql_statement_starts_with (lambda (query keyword) (begin
+	(define trimmed (toUpper (strltrim query)))
+	(define kw_len (strlen keyword))
+	(and
+		(>= (strlen trimmed) kw_len)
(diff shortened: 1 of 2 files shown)


carli2:
sql_statement_starts_with is too hacky and seems to bypass the parser. rejected.

carli2:
Closing this because the patch is not needed in its current form. The intended result-shape bug is not reproducible on current master: empty MySQL and PostgreSQL SELECT-style queries already return an empty row stream rather than affected_rows=0, while DML no-op statements still return affected_rows=0 as expected. The only attempted WITH probe fails because WITH is not currently parsed, which is a separate parser feature, not this result-shape issue. The implementation also classifies statements by raw string prefix, which bypasses the parser and is not the right boundary for this behavior.

PR #273, opened 2026-07-25
Expose uppercase information_schema statistics columns

carli2:
## Summary
- expose MySQL-compatible uppercase aliases for INFORMATION_SCHEMA.STATISTICS index metadata
- keep the existing lowercase column names for compatibility
- add a dbcheck-style regression that selects INDEX_NAME, COLUMN_NAME, and NON_UNIQUE

## Tests
- make
- python3 run_sql_tests.py tests/84_information_schema.yaml
- ./git-pre-commit tests/84_information_schema.yaml

Note: a later full pre-commit attempt was interrupted because I accidentally started full hooks in two worktrees at once; both used the fixed test port/data directory and interfered with each other. The targeted hook above completed successfully before the commit.

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -2107,6 +2107,22 @@ func Init(en scm.Env) {
 					for _, uk := range t.Unique {
 						for seq, col := range uk.Cols {
 							result = append(result, scm.NewSlice([]scm.Scmer{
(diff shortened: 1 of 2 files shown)


carli2:
this is bullshit. mysql is case insensitive. rejected

PR #251, opened 2026-05-20
Top-down query unnesting (Neumann/Kemper) — queryplan rewrite

carli2:
## Summary

Reimplements MemCP's query unnesting subsystem following the Neumann/Kemper papers (NK15 BTW 2015, Neumann BTW 2025). Replaces the old inline scalar-subselect approach with proper top-down decorrelation: every query is untangled into a flat table list + scoped group stages + dependent-join helpers over a domain D, then join-reordered and reassembled into nested selects.

143 commits ahead of master.

## What changed

- `lib/queryplan.scm` — main compiler (untangle_query + build_queryplan)
- `lib/queryplan-tree-ir.scm` (new) — Tree-IR unnester: cclasses union-find, dependent joins, parent-chained UnnestingInfo
- `lib/queryplan-prejoin.scm` — prejoin / keytable materialization for group caches
- `lib/queryplan-debug.scm` (new) — EXPLAIN helpers split out
- `lib/queryplan-legacy-bridges.scm` — isolated compat bridges (slated for removal)
- `storage/` — ORC compute-proxy race fix (recompute sentinel scoped to participating goroutines), MEMORY/CACHE engine, trigger vectorize

## Test status

All 202 test suites green under parallel load (`MEMCP_TEST_JOBS=4` pre-commit hook → "All tests passed, commit allowed"). Test files themselves unchanged vs master — the branch reaches the same test level with the new architecture.

Key fixes verified green this round:
- `66_circular_queryplan` — FAQ §40 cclasses substitution for dependent-domain columns
- `66_vacation_account_group_window_regression` — LEFT JOIN preservation for aggregate-bearing ON predicates (keytable peer placement)
- `32_expr_subselects` — real FAQ §10 once_limit=2 enforcement (was previously passing via a masked exception)
- `73_window_functions` — ORC recompute race under parallel load

## Known follow-up (non-blocking, architecture slimming)

- Remove `legacy_materialized_query_term_binding_ast` + the legacy-bridges module entirely (materialization only for group caches / ORC windows per FAQ §36)
- Decompose `untangle_query` (~4200 lines) and `build_queryplan_inner` (~3200 lines) into recursive IR walkers
- Consolidate the symbol/quote/raw triple-pattern match duplication (391 occurrences)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-07-30
Rebuild query planner around logical operators

carli2:
## Summary
- rebuild query planning around the clean query-block / group-stage / union-block operator model
- keep group-stage semantics logical through untangle/reorder and choose keytable/ORC/carrier details in lowering
- strengthen scalar, EXISTS, IN, derived-table, grouped, window, and DML regression coverage
- fix SQL runner restart/crash handling so crash suites do not hang on expected interrupted SCM calls

## Validation
- make
- python3 tools/lint_scm.py --path lib (exit 0; existing warnings in lib/rdf-parser.scm and lib/psql-parser.scm)
- YAML chunks: tests/00_*.yaml through tests/99_*.yaml via run_sql_tests.py with --fail-fast, split into blocks
- tests/95_crash_recovery.yaml individually with python3 -u

## Notes
- Known noncritical checks remain in tests/83_group_cache_invalidation.yaml for legacy keytable inspection and tests/91_fulltext_like_index.yaml for scan-log speedup instrumentation. Runtime correctness checks in those suites pass.
- Untracked local artifacts were not included: .perf_baseline.json.lock and data.codex-test-backup/.


carli2:
Thanks for getting the clean operator branch to a green CI state. The direction is much better than the large topdown cleanup branch: the logical operator model is smaller, easier to review, and the planner files are much cleaner.

For PR-readiness, I still expect the following before merge:

1. Bring over the missing correctness regressions from PR #251, especially the SQL three-valued NOT IN cases with NULLs on the RHS. Running PR #251 stronger IN-subquery suite against this branch currently shows wrong results for NOT IN with a subquery that returns NULL.

2. Add or port reduced dashboard/ACL regressions from PR #251. This branch passes the central nested IN/EXISTS/HAVING suite, but it still fails some PR #251 dashboard/ACL-style cases, including nested aggregate scalar/dashboard shape and an ACL file-union plan expectation. If the exact PR #251 physical marker is no longer appropriate, replace it with a semantic operator/plan assertion rather than dropping the coverage.

3. Resolve the current noncritical test debt. The legacy keytable inspections in the group-cache suite and the LIKE/fulltext speedup checks currently fail but are marked noncritical. If those physical checks are obsolete under the new operator model, replace them with equivalent semantic IR/lowering/performance gates. If they represent real lost behavior, fix the implementation.

4. Keep the test gates honest. It is fine to change brittle physical-name assertions when the model changes, but the original intent must stay covered: correct NULL semantics, no leaked inner_select markers, no ordinary legacy materialization for derived wrappers, membership-over-UNION avoiding unnecessary duplicate materialization, scalar cardinality behavior, and reusable group-stage carrier semantics.

5. Update the PR description with an explicit scope statement: CI is green and the architecture is the right direction, but this branch is not yet a full behavioral superset of PR #251. List which PR #251 regression classes are covered, which are intentionally replaced by new semantic assertions, and which remain follow-up blockers.

In short: this is the architecture we should continue with, but it should not merge merely because CI is green. It needs to absorb the important correctness/performance lessons from PR #251 without importing that branch complexity.

PR #285, opened 2026-08-01
Remove automatic scan_batch rewrite

carli2:
## Status: Draft / parked

This PR is intentionally parked. The latest measurements do **not** justify merging it as a performance PR.

## Current scope

This branch only removes the automatic `scan_batch` optimizer rewrite while keeping explicit `scan_batch` available as a storage primitive.

## Measurement update

After separating the earlier broad #285 changes, the performance picture is mixed and not conclusive enough for merge:

- ordinary SQL join perf on `origin/master` is already faster than or comparable to this branch
- the harmful automatic batching path is not clearly exercised by normal SQL execution in the tested cases
- earlier measurements that looked better had additional unrelated optimizations mixed in

Real application query A/B sanity checks against an isolated Dynamic-LIKE branch, **without batch changes**, also do not justify a separate speedup PR yet:

| Term class | origin/master median | isolated Dynamic-LIKE median | Result |
| --- | ---: | ---: | --- |
| broad one-character term | 3.632 s | 3.656 s | no speedup |
| highly selective concrete term | 3.221 s | 3.185 s | within noise |
| middle-selectivity prefix term | timeout/hangs | timeout/hangs | unresolved broader planner/storage issue |

## Decision

Do not merge this PR as-is.

Next work should be a controlled benchmark/analysis package:

1. measure candidate optimizations separately from batching
2. only PR isolated optimizations that show repeatable real-query speedup
3. then analyze batching strategies independently:
   - explicit `scan_batch` remains available
   - generic automatic nested-scan-to-`scan_batch` rewrite is suspicious
   - any future batching must be selected by a specific physical planner rule with clear selectivity/cost evidence

## Validation already run on the parked branch

- `MEMCP_TEST_JOBS=1 make test` passed
- `tests/87_scan_batch_optimizer.yaml` passed
- join perf smoke passed, but did not show performance benefit over master


carli2:
Closing this PR intentionally.

The investigation did not produce a mergeable performance change. After separating the earlier mixed scope, the remaining PR only removes the automatic `scan_batch` optimizer rewrite while keeping explicit `scan_batch` available. That cleanup alone is not enough to justify merging, and the measurements do not support presenting this as an optimization.

What we learned:

- Normal SQL execution is already using a different path for the relevant real application query cases.
- The broad one-character search term is correctly planned order-first on the driving table with `ORDER BY ... LIMIT`.
- The highly selective concrete search term is correctly planned candidate-first through the membership candidate.
- The problematic case is the middle-selectivity prefix range, where the planner currently treats the candidate set as selective even though the downstream document join, ACL/reference joins, and final sort make that plan too expensive.
- Isolated Dynamic-LIKE changes did not show a repeatable real-query speedup.
- The older experimental performance branch was slower as a package and should not be ported wholesale.

Conclusion:

The next useful work is not a generic batching rewrite. The next useful work is a focused cost-model/reorder PR for membership candidate stages in ordered limited queries: keep order-first for broad/mid-broad terms, keep candidate-first for truly small candidate sets, and expose enough telemetry in EXPLAIN to make that decision testable.

`scan_batch` should remain available as an explicit storage primitive. We can re-evaluate batching later only as a specific physical planner strategy with clear selectivity/cost evidence, not as a generic nested-scan rewrite.

carli2:
Closed intentionally after measurement: this branch is not a mergeable performance improvement. The next work item is membership-candidate cost model tuning, not generic scan_batch rewriting.

PR #291, opened 2026-08-03
Centralize query shape facts and decorrelate EXISTS

carli2:
## What
- Add central `source_list_shape` / `query_block_shape` fact helpers in `lib/queryplan.scm`.
- Add `shape_fact?` and `shape_has_all?` so support predicates can declare the query shape they need instead of duplicating nested condition cascades.
- Convert the scalar/exists shape support predicates to consume those facts.
- Normalize table-backed `EXISTS` into the existing scalar aggregate group-stage path: logically `MAX(true)` over the inner relation plus `coalesceNil(..., false)` at the expression edge.
- Preserve scalar group-stage domain rows during lowering when `preserve_empty_domain` is set, so left-miss groups survive as domain rows instead of being cleaned up as absent RHS groups.
- Preserve `false` scalar values distinctly from "no scalar row" by using the existing `scalar_first_payload` representation for stage-output/coalesced scalar values that need it.
- Physicalize stage-output sources before building domain seed scans, so preserve-empty domain seeding can operate on already prepared helper relations.
- Rebuild session-dependent group carriers per statement, including transitive dependencies through stage-output sources, so cached plans do not reuse stale helper rows after session variables change.
- Fill scalar-first stages that depend on nested membership stage outputs through a domain-joined input and group on lookup-domain keys. This keeps `NOT IN` miss rows under the left-domain key instead of nullable RHS helper keys.

## Why
The planner had kept `EXISTS` partly as a separate logical marker path. That blocks later Neumann-style optimization because the relation is no longer fully exposed as a joinable/groupable stage after untangling.

This PR moves `EXISTS` closer to the intended architecture: the logical phase emits regular `group-stage(D)`/scalar aggregate structure, and physical helper choices remain in `build_queryplan`. It also fixes correctness holes exposed by deeply nested `EXISTS`/`NOT IN` scalar chains: miss rows for a concrete outer domain now survive long enough to evaluate to `false`/`true` correctly instead of leaking `NULL` or stale cache state.

This is still intentionally bounded. It does not add storage operators and does not include application-specific data or query text.

## Validation
- `make`
- `python3 run_sql_tests.py tests/40_exists_subquery.yaml tests/66_exists_session_var.yaml tests/66_campaign_state_group_regression.yaml tests/66_scalar_subselect_groupby.yaml tests/66_derived_table_limit_scalar.yaml tests/66_vacation_account_group_window_regression.yaml`
- `python3 run_sql_tests.py tests/41_in_subquery.yaml tests/106_subselect_in_clauses.yaml tests/105_deep_nested_in_exists.yaml`

All listed YAML suites are green locally. `tests/105_deep_nested_in_exists.yaml` is now 18/18, including `Mixed depth 4: NOT IN skipping 2 levels to outermost`.

`python3 tools/lint_scm.py` is still not a clean gate on current tree state: it reports the known Scheme warning block (`lib/queryplan.scm`, `lib/rdf-parser.scm`, `lib/psql-parser.scm`). I did not commit broad formatter churn.

## Review Notes
The important architectural change to review is that normal table-backed `EXISTS` no longer uses `exists_probe` in the logical rewrite path. `tests/40_exists_subquery.yaml` now asserts `scalar_aggregate`/`max` and explicitly rejects `exists_probe` in the logical IR checks.

The new domain seed lowering is only enabled for stages that already carry `preserve_empty_domain`; ordinary aggregate cleanup behavior remains unchanged. Stage-output sources in that seed path are physicalized before dataset-row lowering.

Domain seeding now uses only sources referenced by lookup keys, so unrelated LEFT JOIN clauses with still-normalizing subqueries are not pulled into the seed plan.

For session-sensitive group stages, the plan remains reusable, but the physical helper carrier is rebuilt for each statement when its own expressions or transitive stage-output dependencies read session state. This avoids specializing the queryplan cache by session variable values.

## A/B Perf Measurements
Measured serially on 2026-08-04 against copied production-like local data, not the original data directory.

Commits:
- `master`: `7c93c67ba`
- this PR: `171e4bc4e`

Dataset shape used for the application query:
- driver table: 855,496 document-like rows
- file/search table: 57,344 file-like rows
- ACL table: 342 ACL rows
- user/session setup: `SET @fop_user := 342`

YAML A/B, 3 serial runs per suite, wall time includes runner/server startup and is therefore useful only as regression smoke signal:

| Suite | master | PR #291 | Result |
| --- | ---: | ---: | --- |
| `tests/40_exists_subquery.yaml` | 1.333s median, 15/15 | 1.356s median, 15/15 | no meaningful perf movement |
| `tests/66_exists_session_var.yaml` | 1.328s median, 8/8 | 1.398s median, 8/8 | no meaningful perf movement |
| `tests/66_vacation_account_group_window_regression.yaml` | 1.320s median, 2/2 | 1.325s median, 2/2 | no meaningful perf movement |
| `tests/105_deep_nested_in_exists.yaml` | 2.232s median, 13/18 failing | 3.533s median, 18/18 | correctness fixed; timing not directly comparable |

Application full-search query A/B on copied local data, one serial run per term through the MySQL protocol, query output discarded:

| Search term | master SELECT | PR #291 SELECT | Result |
| --- | ---: | ---: | --- |
| `C` | 10.62s | timed out at 130.00s | hard regression |
| `Ca` | 12.80s | not run after `C` timeout | blocked by same path |
| `Car` | 12.74s | not run after `C` timeout | blocked by same path |
| selective term | 22.64s | not run after `C` timeout | blocked by same path |

Plan marker check from `EXPLAIN REORDER`:

| Search term | master | PR #291 |
| --- | --- | --- |
| `C` | 7.49s compile, `exists_probe=18`, `legacy_materialized=0`, `in-candidate=3` | 5.65s compile, `exists_probe=0`, `legacy_materialized=0`, `in-candidate=3` |
| selective term | compile time not captured in the master run, `exists_probe=18`, `legacy_materialized=0`, `in-candidate=4` | 10.45s compile, `exists_probe=0`, `legacy_materialized=0`, `in-candidate=4` |

Conclusion: the architectural normalization works in the logical plan and CI is green, but this PR is not performance-ready for the real application full-search query yet. Removing `exists_probe` without a fast physical replacement currently moves the broad application query onto a much slower execution path. Before undrafting/merging, the normalized presence-only aggregate needs a physical lowering with equivalent early-out/probe behavior or another plan choice that preserves the architectural IR while avoiding the 130s timeout.


carli2:
we won't merge this. it's going into the wrong direction

PR #305, opened 2026-08-05
Develop cost-based document search lowering

carli2:
## Summary

This draft explored cost-based document-search lowering on top of the current operator/RecSet planner. After rebasing onto current `master` with the ordered IN-union RecSet lowering, the branch is **not mergeable**.

The remaining issue is not a minor performance variance: the production-derived document dataview query still fails in final `build_queryplan` on this branch with an unresolved logical stage-output reference during physicalization.

## Current validation

Local checks on the branch after the latest catalog investigation:

- `python3 tools/lint_scm.py --check --path lib/queryplan.scm`
- `make`
- `python3 run_sql_tests.py tests/41_in_subquery.yaml` -> 44/44 passed
- `python3 run_sql_tests.py tests/32_expr_subselects.yaml` -> 43/43 passed
- `python3 run_sql_tests.py tests/119_badge_scalar_aggregates.yaml` -> passed in the earlier rebase validation
- `python3 run_sql_tests.py tests/123_recset.yaml` -> passed in the earlier rebase validation

These targeted tests are not sufficient for merge because the real dataview query still catches a nested stage-catalog/lowering bug that the synthetic tests do not cover.

## Benchmark status

Dataset: production-derived document dataview dataset, local only. PR text intentionally avoids customer/project names.

Current `master` already includes the ordered IN-union RecSet lowering from #293. That is the relevant baseline for this PR.

Local smoke control against current `master` / #293:

| query shape | status | runtime |
| --- | ---: | ---: |
| selective document-search term | 200 | 14.807s cold/single run |
| `EXPLAIN` same term | 200 | 15.08s |
| `EXPLAIN REORDER` same term | 200 | 11.41s |
| `EXPLAIN IR` same term | 200 | 0.77s |

Branch #305 after rebase and catalog fixes:

| query shape | status | runtime |
| --- | ---: | ---: |
| selective document-search term | 500 | fails after ~10.6s |
| `EXPLAIN` same term | 500 | fails after ~12.6s |
| `EXPLAIN REORDER` same term | 200 | ~11.2s |
| `EXPLAIN IR` same term | 200 | ~0.8s |

Failure:

```text
build_queryplan: physicalize stage-output source references unknown stage exists:c86b90081d3d9623 from source __exists_885192772e105435
```

This means the logical/reordered IR still contains the semantic information, but final physical lowering is using a shortened stage catalog in at least one nested grouped/scalar/reference path.

## Assessment

Do **not** merge this PR as a performance package.

Reasoning:

- It is not faster than current `master` / #293; it cannot execute the real benchmark query at all.
- The earlier runtime table in this PR body was stale and predates the #293 baseline.
- The branch demonstrates the right failure area: stage-output catalog lifetime across scalar/presence/grouped lowering. But the fix is incomplete and should not be mixed with performance claims.

Recommended next step: close or keep this PR as a draft investigation only, then start a smaller PR focused solely on the stage-catalog invariant: local `qb_stages` may be reduced for prepare selection, but physical source resolution must always have access to the full logical stage catalog. Add a synthetic regression that mirrors the nested reference/presence/scalar shape before trying cost-based performance changes again.

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -4528,7 +4528,11 @@ PostgreSQL parsers should both lower to the same combined operators.
 			(or (not (nil? (qb_limit input))) (not (nil? (qb_offset input)))))
 			(neumann_fail "build_queryplan" "scalar-first query probe cannot preserve nested ORDER/LIMIT yet")
 			(begin
-				(define raw_probe (cons (quote !begin) (merge (list prepare_exprs (list probe_expr)))))
+				(define raw_probe (cons (quote !begin)
+					(merge (list
+						(lazy_stage_prepare_bindings nested_stages (filter nested_stages group_stage?))
+						prepare_exprs
+						(list probe_expr)))))
 				(if (presence_bool_stage_output_expr? value_expr)
 					(list (quote coalesceNil) raw_probe false)
 					raw_probe))))))
@@ -6250,7 +6254,8 @@ PostgreSQL parsers should both lower to the same combined operators.
 				(list (quote +)
 					(list (quote cadr) (quote old))
 					(list (quote cadr) (quote new))))))
-		(list (quote scan)
+		(define stage_catalog (if (query_block? prepared_input) (query_block_stage_catalog prepared_input) '()))
+		(define scan_plan (list (quote scan)
 			'(session "__memcp_tx")
 			rows_plan
 			(quoted_runtime_list '())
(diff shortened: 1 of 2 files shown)


carli2:
Closing this draft.

Current assessment: this branch should not be merged. CI is failing, and the production-derived document dataview benchmark does not execute on this branch: final build_queryplan still fails with an unresolved stage-output reference during physicalization. Current master with the ordered IN-union RecSet work is the valid baseline and executes the smoke case; this branch does not provide a speedup.

The useful finding from this investigation is narrower: stage catalog lifetime and prepare-set selection need to be fixed as their own architecture PR. Local qb_stages may be reduced for prepare selection, but physical source resolution must still have access to the full logical stage catalog. Separately, the IR -> REORDER jump is far too expensive, around 0.8s to 11-12s on the measured query, and needs a focused compile-time profiling/complexity PR.

Next steps should be split into clean follow-ups:

1. small stage-catalog invariant fix with a regression test for nested reference/presence/scalar stages;
2. separate reorder compile-time profiling and reduction work.

Reason for closing: #305 currently breaks the target query and does not improve performance against current master.

PR #315, opened 2026-08-05
Bound nested scalar plan growth

carli2:
## Summary

- add explicit serialized-plan size guards for existing scalar-subselect regression shapes
- add a depth-4 correlated scalar chain with row-level correctness assertions and a 120 KB plan-size ceiling
- keep the regression matrix in the active two-digit SQL suite so the pre-commit hook and CI execute it

This is a test-only package. It establishes compile/plan-growth limits before the next physical-lowering optimization and does not change engine behavior.

## Validation

- `python3 run_sql_tests.py tests/32_expr_subselects.yaml`: 44/44 passed
- full `./git-pre-commit` suite: passed
- `git diff --check`: passed

## A/B prefix cascade

Both revisions were built independently and run against equivalent copies of the production-scale fixture. Cold is one fresh-process run; warm is the median of three subsequent runs. Times are seconds.

| Prefix | master cold | PR cold | master warm | PR warm |
| --- | ---: | ---: | ---: | ---: |
| C | 2.995 | 3.201 | 0.463 | 0.460 |
| Ca | 3.011 | 3.027 | 0.474 | 0.472 |
| Car | 2.959 | 3.024 | 0.471 | 0.476 |
| Carg | 2.986 | 3.057 | 0.464 | 0.470 |
| Cargl | 2.984 | 2.997 | 0.460 | 0.480 |
| Cargla | 2.964 | 3.009 | 0.458 | 0.471 |
| Carglas | 2.971 | 3.047 | 0.464 | 0.473 |
| Carglass | 3.007 | 3.003 | 0.462 | 0.481 |
| CarglassXXX | 2.981 | 3.000 | 0.472 | 0.472 |
| **Full cascade** | **26.857** | **27.365** | **4.161** | **4.263** |

The test-only diff is performance-neutral within run-to-run variance; no speedup is claimed. Every prefix returned zero rows on both revisions with identical response size and SHA-256 hash. The fixture benchmark therefore validates the same planner/execution path and result stability, while the anonymized SQL regression asserts concrete returned row values.

### DIFF
--- a/tests/32_expr_subselects.yaml
+++ b/tests/32_expr_subselects.yaml
@@ -1,3 +1,18 @@
+# Copyright (C) 2026  Carl-Philip Haensch
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


carli2:
useless

PR #332, opened 2026-08-06
Reduce indexed stage lowering compile time

carli2:
## Summary

- cache the stage dependency graph in the existing indexed lowering catalog
- reuse the indexed catalog while rewriting and lowering nested query blocks instead of repeatedly merging the global stage list into every owner
- keep only the local dependency closure in stage materialization metadata
- deduplicate typed stage lists by logical stage ID instead of deep AST equality
- add anonymous correctness and cold-planner regression coverage to the CI-discovered two-digit suite `tests/96_scalar_subselect_patterns.yaml`
- run that compile-budget suite in the hook's isolated phase so concurrent rebuild suites cannot invalidate wall-clock measurements

The earlier direct scalar-aggregate lowering attempt was fully reverted in `23863d855`. This replacement preserves the existing logical and physical plan; it only reduces compiler metadata and list-processing work.

## Root cause

Wide projections with independent nested scalar aggregate and presence probes produce a large, structurally diverse stage catalog. Recipe emission repeatedly rebuilt that catalog, copied the complete list into individual stage facts, recalculated the dependency graph, and deduplicated stage lists through deep AST equality. The repeated immutable list and AST work dominated compile time.

The indexed catalog now owns the immutable dependency graph. Global ID resolution stays in that catalog, while each materialization receives only its local dependency closure. Lists known to contain stages use logical IDs for deduplication.

## Test-first regression

The anonymous test contains 23 structurally distinct nested projections and enforces:

- `max_planner_time_ms: 1000`, checked from cold `EXPLAIN COMPILE` `planner_total_ns`
- `max_plan_size: 400000`
- `LIMIT 0`, so the test measures compilation without executing the nested subqueries

On unchanged `origin/master` (`57c2880c7`) the test failed five out of five runs: 1075.5-1113.7 ms. On this branch it passed five out of five runs. The test is in `tests/96_scalar_subselect_patterns.yaml`, which is selected by the normal `tests/[0-9][0-9]_*.yaml` CI discovery.

## A/B measurement

Same anonymous query and host, five `EXPLAIN COMPILE` requests per binary, detached master worktree versus PR worktree:

| Metric | master median | PR median | Change |
| --- | ---: | ---: | ---: |
| planner total | 1090.1 ms | 967.4 ms | -11.3% / 1.13x faster |
| recipe emit | 961.6 ms | 846.1 ms | -12.0% / 1.14x faster |

Plan shape is identical in both builds:

- logical nodes: 36,352
- plan nodes: 20,811
- plan bytes: 220,390
- group stages: 69
- scans: 205
- ordered scans: 24

## Validation

- `tests/96_scalar_subselect_patterns.yaml`: 23/23, compile-budget case repeated five times
- `tests/118_manual_trace_sql_regressions.yaml`: 20/20
- full normal pre-commit hook: all suites passed in 319 seconds (5:19)
- `tests/14_order_limit.yaml`: 19/19 in 1.89 seconds with the normal binary
- `python3 tools/lint_scm.py` run; its remaining warnings are present on unchanged master

The local coverage-instrumented `make test` run hit the pre-existing 5-second runtime gate in `tests/14_order_limit.yaml` at 8.13 seconds under host load; the same suite passed immediately with the normal binary. This PR remains draft until GitHub's coverage CI is green and its total job duration confirms that the previous 9-minute regression is gone.


carli2:
Superseded by #335, which implements inherited indexed lowering catalogs more generally. The remaining experimental changes show no measurable improvement over current master.

PR #380, opened 2026-08-08
Reuse query-local scalar probe results

carli2:
## Summary

- add a thread-safe `memoize` runtime wrapper for pure functions, keyed by structurally equal argument tuples
- keep small caches as linear assoc pairs and promote at the existing five-entry assoc-array threshold to avoid eager map allocation
- wrap generated scalar query-probe recipes so repeated correlation keys reuse completed lookup results within one query execution
- preserve recursion and panic behavior: misses run outside the cache lock, and failed evaluations are not cached

## A/B measurements

Base: `03a3bdc2b` (current master at measurement time)  
PR: `9f293a0a3`  
Same production-derived data snapshot, same host, sequential processes, TracePrint enabled. Each prefix run used one warmup followed by three measured executions. Values are median wall time in seconds; lower is better. Prefix contents are intentionally anonymized.

| Prefix stage | Master | PR | Delta |
|---:|---:|---:|---:|
| 1 char | 25.140 | 22.827 | -9.2% |
| 2 chars | 5.935 | 5.572 | -6.1% |
| 3 chars | 4.035 | 3.745 | -7.2% |
| 4 chars | 3.871 | 3.678 | -5.0% |
| 5 chars | 3.847 | 3.744 | -2.7% |
| 6 chars | 3.751 | 3.663 | -2.3% |
| 7 chars | 3.860 | 3.637 | -5.8% |
| 8 chars | 3.893 | 3.736 | -4.0% |
| longest | 1.030 | 1.002 | -2.8% |

The first broad warmup had one branch outlier (`35.423 s` versus `25.390 s`). A separate fresh-process repeat did not reproduce it: master `26.495 s`, PR `26.216 s`. The three measured broad samples were stable: master `24.562..25.843 s`, PR `22.238..23.300 s`.

The production-derived wide detail query was also measured five times per fresh process through the MySQL frontend:

| Variant | Samples (s) | Median |
|---|---|---:|
| Master | 8.383, 8.033, 7.707, 7.701, 7.697 | 7.707 |
| PR | 7.578, 7.536, 7.405, 7.379, 7.279 | 7.405 (-3.9%) |

Both variants returned 72 rows, 13,272 bytes, with identical SHA-256 output.

## Verification

- `go test -race ./scm -run '^TestMemoizedFunction'`
- `go test ./scm`
- targeted correlated-probe and deeply nested correlation SQL suites
- full `make test` after merging current master

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -5400,8 +5400,9 @@ pass correlation keys to it instead of copying the complete recipe per field. */
 			(list
 				(quote define)
 				(symbol (scalar_query_probe_recipe_key stage requested_col))
-				(list (quote lambda) params
-					(lower_scalar_first_query_probe_expr_using stage value_expr keys params nested_stages prepare_stages))))
+				(list (quote memoize)
+					(list (quote lambda) params
+						(lower_scalar_first_query_probe_expr_using stage value_expr keys params nested_stages prepare_stages)))))
 		_ (neumann_fail "build_queryplan" "malformed scalar query probe recipe plan"))))
 
 (define scalar_query_probe_recipe_bindings (lambda (plans)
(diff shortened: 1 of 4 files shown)


carli2:
Closing this experiment: query-local memoization masks redundant scalar-probe work but adds a separate stateful cache with memory and parallel-scan scaling risks. The replacement will address the decorrelated join orientation and regular scan path directly.

PR #394, opened 2026-08-09
Fix nested aggregate stage lowering

carli2:
## Problem

A no-FROM projection composed from sibling scalar aggregates failed when one aggregate contained another correlated aggregate. Probe rewriting reduced the outer block to one physical source, but dependency stages remained attached to `qb_stages`; the single-source core therefore rejected the otherwise prepared plan.

The same minimal query exposed a second correctness issue: equal-precedence additive operators were parsed right-associatively, so `a - b - c` became `a - (b - c)`.

## Fix

- clear logical stages from the already prepared physical core block while retaining the lazy stage catalog used for carrier preparation
- parse additive operator chains once and fold them left-to-right in both SQL frontends
- add anonymous session-sensitive nested aggregate regressions with a 1 s runtime bound
- add a direct left-associativity regression for mixed `+` and `-`

## A/B measurements

| Measurement | origin/master | branch |
|---|---:|---:|
| Nested aggregate target suite | 21/23, 2 HTTP 500 | 23/23 |
| Target suite wall time | 1.467 s | 1.417 s |
| Downstream phase-1 manual checks | 7/19 pass | 12/19 pass |
| SQL compiler errors during manual checks | 61 | 0 |
| Unrelated ordered-pagination suite, release binary | 19/19, 1.820 s | 19/19, 1.813 s |

The remaining downstream failures are UI/form synchronization timeouts; the engine error table remained empty.

## Validation

- `tests/planner/aggregates/composed-scalar-aggregate-shapes.yaml`: 23/23
- `tests/sql/expressions/basic-sql.yaml`: 27/27
- `tests/sql/compatibility/psql-parser.yaml`: 52/52
- `tests/sql/expressions/date-functions.yaml`: 65/65
- downstream manual setup/init checks and phase 1 completed, with 12/19 phase-1 checks passing

`make test` traversed the complete suite and failed only two 5 s ordered-pagination performance assertions while using the Go coverage binary (9.95 s and 5.61 s, both correct results). Rebuilding without coverage makes the suite pass 19/19 in 1.813 s; the equivalent master binary takes 1.820 s.

### DIFF
--- a/lib/psql-parser.scm
+++ b/lib/psql-parser.scm
@@ -107,6 +107,15 @@ arithmetic; leave expressions containing columns or functions untouched. */
 		'("date_add" value unit) '('date_add acc value unit)
 		'("date_sub" value unit) '('date_sub acc value unit))))
 
+(define psql_comparison_expr (lambda (op a b)
+	(match op
+		"eq" '((quote equal??) a b)
+		"ne" '((quote not) '((quote equal?) a b))
+		"le" '((quote <=) a b)
+		"ge" '((quote >=) a b)
+		"lt" '((quote <) a b)
+		"gt" '((quote >) a b))))
+
 (define psql_type (parser (or
 	(parser '((atom "character" true) (atom "varying" true)) "varying")
 	(parser '((atom "double" true) (atom "precision" true)) "double")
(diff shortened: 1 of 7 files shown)


carli2:
CI follow-up: reduced additive-tail parser backtracking after the hard compile-time check missed its 1 s budget by 47 ms. Forced 80 KB query A/B on release binaries: origin/master 654.7 ms total (545.1 ms parse), branch 645.7 ms total (547.2 ms parse). The original hard-limit suite now passes locally.

carli2:
CI follow-up after the 1 s compile-budget failure:

- Root cause in the failing shape: parsing dominated (`854.6 ms` parse vs `121.8 ms` planner), with substantial GC pressure from memoizing a purely linear additive tail loop.
- The additive tail now uses go-packrat's existing `NoMemo` path; recursive expression parsers remain memoized.
- Local 80,918-byte query, repeated on one instance: parse median improved from about `547 ms` to `470 ms` (`-14%`); HTTP median is `565 ms`.
- SQL/PSQL parser suites, aggregate regressions, `go test ./...`, and the complete repository hook passed. The full hook was run with `MEMCP_TEST_JOBS=1` because `transactions.yaml` interferes with other suites on a shared parallel instance; isolated it passes 171/171.

carli2:
Second CI follow-up:

- CI still measured `875.2 ms` parsing and `972.0 ms` internal compile (`1026.7 ms` HTTP).
- Consolidated eight comparison branches that all reparsed the same LHS into one operator-dispatch branch in both SQL grammars. Semantics remain covered for `=`, `==`, `<>`, `!=`, `<`, `<=`, `>`, `>=`.
- Same local 80,918-byte query: parse median `470 ms -> 408 ms` (`-13%`), HTTP median `565 ms -> 493 ms`. Relative to the initial branch measurement (`547 ms` parse), parsing is now about `25%` faster.
- Full serialized SQL hook passed after the change.

Separate master issue observed: `go test ./...` times out after 10 minutes in `TestManualRepartitionForwardsConcurrentInserts` on a repartition/insert lock cycle; this patch does not touch storage.

PR #425, opened 2026-08-11
Optimize correlated EXISTS UNION probes

carli2:
## What changed

- Preserve compatible correlated `EXISTS (... UNION ...)` branches as one logical union-backed presence stage.
- Lower selective scalar presence probes to a lazy `OR` chain of bounded `scan_exists` operations, so no UNION result is materialized.
- Keep broad filtering on the existing RecSet union/project semijoin path.
- Fall back to the existing branch-wise lowering unless every branch has one correlation key, one lookup key, no residual outer references, and one base source plus preparable stage helpers.
- Add anonymous SQL regression coverage for both selective bounded probes and broad RecSet semijoins.

## Why

The previous branch-wise rewrite expanded a trace-derived 12-branch correlated presence query into independent group-cache stages. That increased generated plan size, compile work, and cache initialization. The new logical stage preserves the common UNION structure until physical lowering can choose the consumer-specific plan.

## A/B against `origin/master`

Five cold `EXPLAIN COMPILE` samples were collected from isolated servers using the same anonymous 12-branch query.

| Metric | master | PR | Delta |
| --- | ---: | ---: | ---: |
| Median compile | 371.716 ms | 348.848 ms | -6.15% |
| Median planner | 349.911 ms | 326.508 ms | -6.69% |
| Physical plan nodes | 24,163 | 15,087 | -37.56% |
| Serialized plan bytes | 257,589 | 185,470 | -28.00% |
| Scans | 234 | 157 | -32.91% |
| Group caches | 49 | 24 | -51.02% |
| Bounded existence scans | 0 | 12 | +12 |

Runtime on the same fixture:

| Metric | master | PR | Delta |
| --- | ---: | ---: | ---: |
| Cold query | 413.053 ms | 375.041 ms | -9.2% |
| Warm median | 6.194 ms | 4.822 ms | -22.1% |

The trade-off is larger logical IR before physical lowering: 24,431 to 34,251 nodes (+40.20%), and 36 to 49 reported group stages. This comes from retaining branch-local nested stage references. The physical plan is substantially smaller, but further logical compaction remains useful.

## Validation

- `python3 run_sql_tests.py tests/planner/subqueries/traced-union-scalar-probes.yaml`: 5/5 passed.
- Scheme formatter executed; its check mode still reports the repository's pre-existing negative-depth warnings.
- The application manual suite passed setup and initialization and reached its parallel detail-query phase. The next blocker is separate: concurrent readers can observe a canonical group-cache table before an expected aggregate column exists.
- Full local `make test` is blocked by `tests/execution/operators/range-scan-coverage.yaml`: the scan-statistics aggregation times out. The same 84/85 failure reproduces on the unchanged `origin/master` binary at the PR base commit, so it is not introduced by this patch.


carli2:
Superseded by a clean implementation from current master. The measured EXISTS/UNION opportunity and regression fixture remain valuable, but this branch carries obsolete join-reorder/lowering history and collapses independent UNION branches into a nested binary OR shape, conflicting with the structural tree invariants now established by #428/#438. Replacement branch: optimize/exists-union-structural.

PR #430, opened 2026-08-12
Preserve logical join trees through physical lowering

carli2:
## What changed

- keep `qb_sources` as the semantic alias/source catalog instead of rewriting it into physical order
- make `join_plan` the single owner of physical join order
- consume the logical join tree recursively during physical lowering, preserving bushy subtrees and LEFT OUTER boundaries
- attach normalized WHERE/INNER-ON predicates to the lowest selected join node once, after join ordering
- retain INNER-ON/OUTER-ON provenance and outer-barrier ownership on logical join predicates
- express constant-scalar-first preparation by updating `join_plan`, not by reordering the source catalog
- remove the cursor-based tree flattening helpers and the separate ordered-join lowering path
- keep ROW_NUMBER lowering inputs explicit and preserve alias-scoped `$update` callbacks for multi-table DML

## Why

The previous physical preparation produced a join tree and then copied its leaf order back into `qb_sources`. Other lowerers subsequently depended on that second, implicit representation and flattened bushy trees into a leaf cursor. This lost join-node and null-extension boundaries and prevented later physical predicate placement from seeing the selected tree structure.

After this change, source lookup and traversal order have separate, explicit owners: `qb_sources` resolves aliases, while `join_plan` drives lowering. Logical predicates remain attached to the chosen join tree with their semantic provenance. This is the architecture prerequisite for a later selection-pushdown change; this PR intentionally does **not** implement selection pushdown.

## Validation

- full local hook-equivalent suite: passed (`./git-pre-commit`)
- unchanged 200,000-row multi-shard ORDER/OFFSET timing fixture: passed
- SQL time-limit guard: 15/15
- SQL runner contract: 4/4
- TPC-H relational query shapes: 22/22
- focused range scan, outer join, scalar/window and multi-table DML suites: all passed
- Scheme tests validate predicate uniqueness/placement and inspect the emitted physical scan tree, including a nested LEFT OUTER boundary

No guard, time limit, CI configuration, or final SQL fixture differs from `master`.

## A/B measurement

Badge scalar aggregate query, same populated database, normal priority, HTTP wall time:

| | cold | warm samples |
|---|---:|---:|
| `master` | 137.3 ms | 45.5, 41.7, 43.7, 43.2 ms |
| this branch | 138.7 ms | 44.0, 37.7, 38.8, 38.0 ms |

Correct results were identical. The branch had initially regressed to roughly 550–590 ms when scalar-stage preparation still reordered `qb_sources`; moving that choice into `join_plan` removed the regression and restored comparable performance.

## CI timing fix

The first CI run exposed the existing 15-way star compile gate at 5363 ms / 5000 ms. The implementation now proves the connected-subgraph budget overflow from vertex degree (`2^d`) before enumeration and selects the same `linearized-dp` fallback without materializing more than 10,000 subsets. The unchanged regression then passed locally; the direct EXPLAIN measured about 30 ms. The full hook passed again after this fix.


carli2:
Follow-up found while rebasing #433 onto the direct join-tree architecture:

- `join_plan` does not currently retain predicates local to the first leaf. `join_order_tree_with_predicates` attaches only crossing predicates to join nodes, while `join-leaf` carries only the alias. A local condition stored on the first source therefore still depends on the physical lowerer rereading `source_join_expr`.
- `join_optimizer_tree_without_aliases` returns the surviving child directly when the other child is removed and drops the removed node predicates. Predicates that remain semantically required need to be transferred to the surviving leaf/node or an explicit residual.

I verified this by removing the `source_join_expr` fallback in #433 and deriving ON/nullable ownership exclusively from the tree. `tests/planner/subqueries/in-subquery.yaml` then regresses `Derived LEFT JOIN keeps projected values beside nested permission probes` from 2 rows to 6 rows (duplicated matches plus an incorrectly preserved NULL row), while the existing ON-before-null-extension and nullable-WHERE tests remain green. Restoring the catalog fallback hides the loss again.

Suggested architecture: let logical leaves own local predicates with provenance (`where`, `inner-on`, `outer-on`), and make tree pruning return/reattach predicates that are still evaluable after alias removal. Then physical lowering can consume leaf selection separately from pre-null-extension ON and use `qb_sources` only for alias/relation resolution.

PR #440, opened 2026-08-12
Optimize conditional zero-one aggregates

carli2:
## Summary

- normalize `SUM(CASE WHEN predicate THEN 0 ELSE 1 END)` and the inverse 1/0 form to a logical conditional count before unnesting
- preserve SQL three-valued and empty-input semantics by computing either `true_count` or `row_count - true_count`, returning `NULL` when `row_count = 0`
- lower cache-column materialization to the existing filtered `scan` with constant payload `1`; no new physical operator
- add anonymous YAML regressions for `FALSE`/`UNKNOWN`, empty scalar SUM, inverse 1/0 form, plan shape, and incremental invalidation

## Plan change

Master materializes the conditional aggregate column by scanning every row and reducing `CASE ... 0/1` through `sql_sum_reduce`. This branch keeps total count and true count in the same cold group initializer, uses primitive `+`, and materializes the conditional cache column through a predicate-filtered scan. The final 0/1 SUM identity is reconstructed from the two counts.

## A/B measurements

Same host, fresh MemCP processes and data directories, identical synthetic tables. Query shape: 500 groups over 50,000 rows, with 0.2% predicate selectivity. Three fresh table names avoid group-cache reuse between samples.

| phase | master | branch | delta |
|---|---:|---:|---:|
| cold query / group-cache creation, median | 488.7 ms | 611.2 ms | +25.1% |
| first conditional aggregate cache read, median | 166.9 ms | 3.0 ms | -98.2% / 55.6x faster |
| subsequent warm read, median | 2.63 ms | 2.71 ms | +3.0% (noise range) |

Raw samples in seconds:

- master cold: `0.499208`, `0.481435`, `0.488743`
- branch cold: `0.605851`, `0.626428`, `0.611164`
- master first conditional read: `0.168433`, `0.166868`, `0.163773`
- branch first conditional read: `0.003010`, `0.003354`, `0.003031`

The intentional trade-off is a roughly 25% slower first group-cache construction because the plan maintains total and true counts. It removes the later full-table CASE/SUM materialization and makes the first selective aggregate lookup roughly 50x faster. Repeated cache reads are unchanged within measurement noise.

## Validation

- `python3 tools/lint_scm.py --check` (passes with the repository's existing depth warnings)
- `python3 run_sql_tests.py tests/planner/aggregates/incremental-aggregates.yaml`: 115/115 passed
- `python3 run_sql_tests.py tests/execution/concurrency/rebuild-concurrent.yaml`: 16/16 passed in 1.1s
- full `make test` progressed through all displayed suites without a failure, then stalled in the shared-process run of `rebuild-concurrent.yaml` with old requests waiting on table locks; stopped after >8 minutes. The same suite passed immediately in isolation as noted above. CI should validate in its clean environment.

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -346,6 +346,53 @@ PostgreSQL parsers should both lower to the same combined operators.
 		(make_query_block schema sources fields where group having order limit offset '() '() '())
 		_ query)))
 
+/* SUM(CASE WHEN predicate THEN 0 ELSE 1 END) is a conditional count, but
+SQL SUM still returns NULL for an empty input. Keep that distinction explicit
+in the logical expression and lower the per-row marker to a primitive IF only
+when aggregate input is emitted. In particular, FALSE and UNKNOWN both select
+the ELSE arm; rewriting this as (not predicate) would be NULL-incorrect. */
+(define sql_conditional_count_result (lambda (row_count true_count count_when_true)
+	(if (equal? row_count 0)
+		nil
+		(if count_when_true true_count (- row_count true_count)))))
+
+(define filtered_count (lambda (predicate)
+	(if predicate 1 0)))
+
+(define conditional_count_aggregate_rewrite (lambda (agg_expr agg_reduce agg_neutral)
+	(if (and (equal? agg_reduce (quote sql_sum_reduce)) (nil? agg_neutral))
+		(match agg_expr
+			((symbol if) predicate 0 1) (list predicate false)
+			((quote if) predicate 0 1) (list predicate false)
+			((symbol if) predicate 1 0) (list predicate true)
+			((quote if) predicate 1 0) (list predicate true)
+			_ nil)
+		nil)))
+
+(define normalize_conditional_count_aggregates (lambda (expr)
+	(match expr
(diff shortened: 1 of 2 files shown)


carli2:
EUR manual validation against this branch binary (`TracePrint=true`, MySQL port 3307, anonymous details only):

- passed setup, initialization, first-user, all 12 phase-1 tests, import, item, invoice/offer/billing, PDF, menu, document actions and document links
- stopped in the later document-merge UI test: the second uploaded-file link did not become visible within the Playwright 30s locator timeout
- trace evidence does not indicate an SQL failure: both uploaded file IDs existed, and filename/size reads completed in roughly 0.2-0.25ms; surrounding badge/ACL reads completed in 11-16ms
- after those successful reads, no SQL request was made for the remaining 30s before the browser assertion failed
- `system_statistic.errors` remained empty

Conclusion: EUR is not fully green yet, but this run reached the document-merge stage. The observed blocker is after successful database reads in the upload/UI refresh path, not conditional aggregate correctness or SQL latency.

carli2:
it's a special case, not an improvement to the general queryplan builder's decision making.

PR #529, opened 2026-08-19
perf: recursive RecSet-vs-keytable decider for boolean scalar-first probes

carli2:
## Summary

PR #527 added a RecSet option for `scalar_single`-purpose probes, but its
eligibility check was itself a hardcoded, exists-only pattern matcher, and
it never fired for the production shape it was meant to help: a
`COALESCE((SELECT bool_expr FROM standort WHERE standort.ID = dokument.standort
LIMIT 1), FALSE)` outer probe whose `bool_expr` nests a further correlated
`EXISTS` against `userMandant`. That stage is consumed as a `qb_sources`
JOIN partner in the driving query, not through the `scalar_first_probe`
marker path #527 targeted, so #527's cost-gated eligibility check was never
even reached for it -- traced and confirmed empirically this session.

This PR replaces the hardcoded matcher with a genuinely recursive decision,
per the discussion in this repo: "der recset compiler ... zuerst die
innerste plan-entscheidung (recset oder scan), dann immer weiter nach
außen".

- `stage_boolean_shaped?` proves a stage's realized value is boolean by
  recursing through its dependency graph (`stage_dependency_graph`/
  `stage_direct_deps`), so an arbitrarily deep `COALESCE(nested EXISTS)`
  chain is recognized just as well as a single-level `EXISTS`.
- `stage_recset_domain_eligible?` (generalizing the old exists-only
  `exists_recset_stage?`) accordingly covers any boolean-shaped
  `scalar_single` or `exists` stage, not only presence-only exists stages
  -- gated by a single-real-source domain-shape check, since only the
  where-term consumption path is proven safe for a multi-source domain.
- `exists_recset_project_join_expr` (the one function both the where-term
  and, previously, a value-embedding path routed through) now builds the
  RecSet from the stage's own **prepared group-cache** (reusing
  `lower_group_stage_prepare_using`) instead of re-deriving the stage's raw
  condition against its domain table. This is what makes it correct for
  arbitrarily nested dependency stages without re-implementing their own
  preparation. `recset_project_join_expr_for_membership`'s old single-key
  inline special case (a second, un-generalized recset_project_join
  construction) is gone -- every key count now goes through this one
  function.
- `group_stage_with_stage_catalog` stamps the discovering query block's own
  stages list onto a stage embedded in a `driver_membership_probe` marker,
  so a `qb_sources`-based nested dependency inside that stage's own domain
  stays resolvable once the stage is consumed far from where reorder
  discovered it.
- `scan_recset` now also accepts a RecSet (not just a table) as its input
  (Go side, `storage/recset.go`), so a cheap selective filter and a more
  expensive correlated condition can compose as two RecSet scans (a
  logical AND) instead of re-scanning the full driving table for each --
  the shape needed for "narrow 800k dokument rows to a mandant's 30k first,
  then evaluate the expensive ACL check only over that mask."

### Also fixed: a real cost-model bug in `storage/recset.go`

Found while verifying this change end-to-end. `projectJoinKeysPart`'s
dense/sparse heuristic sized purely off the *target* table's row count
(`keys.count()*64 >= t.CountEstimate()`), with no regard for whether an
index for the join columns already existed. For a small key set matched
against a large, previously-unindexed driving table, this always picked
the sparse (indexed) path -- forcing a one-off index build that cost far
more than the linear scan it was meant to avoid (observed: 20s for a
single query at 200k rows). It now falls back to the dense (scan) path
unless a matching index already exists (`hasEqualityIndexPrefix`).

## A/B measurement

200k-row driving table (`doc`), 20-row `standort` domain, 1-row
`userMandant`, single qualifying correlation key, cold cache / no
pre-existing index (fresh data dir each run) -- reproducing the
`COALESCE((SELECT EXISTS(...) FROM standort WHERE standort.ID = doc.standort
LIMIT 1), FALSE)` shape:

| | before (master, incl. #527) | after |
|---|---|---|
| `SELECT COUNT(*) FROM doc WHERE COALESCE(...)` | 0.868s | 0.074-0.094s |

~10-12x. `EXPLAIN` confirms the plan shape: before, `doc`'s driving scan
does a per-row keytable point-lookup into the standort presence cache;
after, `doc`'s driving scan is replaced with `recset_project_join` over a
`scan_recset` of that same cache, filtered to the qualifying rows.

Correctness verified: `COUNT(*)` matches a plain `EXISTS(...)` reference
query (10000, matching the one qualifying `standort`'s row count) and
returns 0 for a non-matching session key.

## Test plan / regression status

- [x] `lib/test.scm` unit tests: 1228/1230, matching clean `origin/master`
      exactly (2 pre-existing, unrelated failures: `_mut merge_unique on
      fresh first arg` and `optimizer: dynamic reducer keeps merge
      validation before reducer lookup`)
- [x] Manual EXPLAIN + row-count correctness check for the target
      COALESCE(nested EXISTS) shape, at 200k rows, cold and warm
- [x] A/B wall-clock measurement (above)
- [x] Every SQL test file `make test` initially flagged, re-verified
      file-by-file against a genuinely clean `origin/master` checkout
      (built and run from its own directory -- an earlier verification
      pass in this PR's history incorrectly ran the master *binary* from
      within this worktree, so it loaded this branch's `lib/queryplan.scm`
      instead of stock master's; that false "no regressions" claim has
      been corrected). All are now fully green and match clean master,
      **except one remaining, known regression**:

  **`tests/integration/query-shapes/application-read-models.yaml`**, "Data
  view preserves projected filters through sorter joins and pagination":
  a boolean column backed by a nested-dependency `EXISTS`
  (`COALESCE(nested EXISTS)`-shaped, the same family this PR targets) that
  is both projected from a derived table and re-filtered in the enclosing
  query, breaks specifically when that enclosing derived table also has an
  additional `LEFT JOIN` partner. Extensively isolated (the group-cache
  contents and the `recset_key_index` storage primitive were both verified
  correct in isolation via raw `scm` eval against the failing query's own
  data) but not root-caused -- the bug is in how the compiled query
  evaluates the marker specifically in that shape, not in the underlying
  data or primitives. A `recset_key_index`-based value-returning lowering
  was attempted for this consumption path and reverted after it reproduced
  the same wrong-row result *and* broke two other, previously-working
  tests. `stage_recset_domain_eligible?`'s domain-shape guard cannot
  distinguish this specific shape from the many structurally similar ones
  fixed in this PR, since that distinction depends on how the stage ends
  up being consumed downstream, which isn't known at eligibility-check
  time. See commit f2475d504 for a fuller writeup of what was ruled out.

- [ ] Full `make test` suite: not completed locally within this session --
      per the repo maintainer, CI covers this.

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -7340,21 +7340,38 @@ prepared cache instead. */
 				(if (source_is_base_table? real_src) real_src nil))
 			nil))))
 
-/* No domain-shape restriction here: exists_recset_project_join_expr builds
-the recset from the stage's own prepared group-cache, not its raw domain, so
-this is agnostic to whether gs_input is a base table, a query-block wrapping
-one, or a union -- lower_group_stage_prepare_using already knows how to
-materialize any of those, nested dependencies included. */
+/* exists_recset_project_join_expr builds the recset from the stage's own
+prepared group-cache rather than its raw domain, so in principle this is
+agnostic to whether gs_input is a base table, a query-block wrapping one, or
+a union. In practice the where-term consumption path
+(recset_project_join_expr_for_membership) is the only one proven safe for a
+multi-source domain (e.g. an EXISTS body that itself joins two tables); the
+value-embedding path (lower_driver_membership_probe_expr's scan_exists
+fallback, used whenever the where-term path declines) still needs a single
+concrete table to scan and fails hard otherwise -- so this keeps the
+original single-real-source restriction. A stage with residual outer refs
+(btw2025_accessing_after_simple) reaches further out than its own domain for
+a value that isn't local to it or its declared lookup-keys --
+exists_recset_project_join_expr's stage_catalog (scoped to the discovering
+block's own stages) has no way to resolve that, so those stages are
+excluded here rather than crashing during preparation. */
 (define stage_recset_domain_eligible? (lambda (graph stage)
 	(if (not (group_stage? stage))
 		false
 		(begin
 			(define lookup_keys (qassoc_get (gs_facts stage) (quote lookup-keys) '()))
+			(define input (gs_input stage))
 			(and
+				(not (stage_has_residual_outer_refs? stage))
 				(recset_probe_stage_shape? stage)
 				(stage_boolean_shaped? graph stage nil)
 				(not (empty_list? lookup_keys))
-				(equal? (count lookup_keys) (count (gs_keys stage))))))))
+				(equal? (count lookup_keys) (count (gs_keys stage)))
+				(or (not (nil? (recset_domain_source input)))
+					(and (union_block? input)
+						(reduce (union_branches input) (lambda (supported branch)
+							(and supported (candidate_recset_branch_supported? branch)))
+							true))))))))
 
 (define recset_domain_stage_output_source? (lambda (stages src)
 	(and (stage_output_relation? (source_relation src))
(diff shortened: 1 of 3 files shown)


carli2:
The claim that master's failures are "pre-existing" is wrong. You maybe used your PR's lib/ folder with master binary which leads to having the same queryplanner bugs as in your branch

PR #534, opened 2026-08-20
perf: make correlated-subquery probes actually reachable and fast (real fix, not another layer)

carli2:
## Summary

The user's own diagnosis, translated: the query-plan compiler (~18k
lines) had drifted into thinking in special-case corner-cases instead of
the originally-designed operator logic (query / union / group, with
parameterization), and that structural mess is why three prior PRs
(#522, #523, #527/#529) each added a new physical strategy for realizing
a correlated subquery's value, targeting the same real production
`omnestum` `dokument` pagination COUNT query, and each time the new
strategy never actually fired for the real query. The user's mandate for
this PR: a real, mechanically-verified improvement, not more code with
zero reachability.

**Target query shape**: a correlated scalar subquery embedded in a WHERE
clause (not a projected field), with a compound `CASE/OR/EXISTS/COALESCE`
value, correlated against a large driving table with no `LIMIT` — the
exact shape of the real production ACL check that every prior attempt
missed.

### The three-phase plan (as designed)

1. **Test first** (`acl-boolean-membership-scaling.yaml`): a permanent
   regression modeling the real predicate shape at 100k driving rows,
   written to fail on current master before any fix.
2. **Generalize `stage_boolean_shaped?`** to recognize compound boolean
   trees (CASE/AND/OR/COALESCE chains), not just the two narrow shapes it
   previously matched.
3. **Unconditional probe promotion**: `stage_probe_allowed_in_context?`
   previously gated whether a correlated stage could ever become a probe
   marker (the only thing any physical carrier — keytable, RecSet, or the
   safe materialized-query-probe fallback — ever looks at) on a
   driving-context row-count heuristic (`<1000` rows), which structurally
   excluded exactly the large-table/no-LIMIT correlated subqueries that
   dominate real production cost.

### A real regression, found and fixed along the way

Phase 3 (unconditional promotion) introduced a genuine, severe
regression, caught by the pre-commit hook and — after I initially
misdiagnosed it as pre-existing flakiness by comparing against my own
already-modified branch commit instead of real master — correctly
challenged by the user to re-verify properly. Bisecting phase-by-phase
against a clean `origin/master` build confirmed it: Phase 2's blanket
removal of the row-count gate also silently broke a second, unrelated
safety mechanism — `query_block_with_scalar_first_probes_using_graph`'s
`prelimit_aliases`/`unbounded_probe_aliases` split, which stops a
"prelimit source" (a stage referenced by WHERE/ORDER that gates which
rows satisfy an active `LIMIT`) from becoming a probe marker unless it's
*also* safe to promote independent of the limit. A promoted stage's
physical carrier has no way to know it only needs to satisfy a handful
of driving rows before a `LIMIT` is met, so promoting a prelimit source
can force an eager, whole-domain build where the native-limit ordered
scan's lazy, early-terminating per-row check would have stopped after a
few rows.

Measured on `tests/planner/order-window/order-limit.yaml`'s "Scalable
bounded derived read model applies nested permission before native
limit" case (20k driving rows, a correlated EXISTS probe gating an
`ORDER BY ... LIMIT 2`): **0.045–0.12s on master, 2.27–3.4s after Phase
2** — a ~45x regression, identical regardless of anything layered on top
in later phases, confirming it dates to Phase 2 specifically.

Fixed by restoring the original row-count-based safety net, but gating
it on whether the query block actually has an active `LIMIT`/`OFFSET`
(`query_limit_active?`): unconditional promotion is kept when it
doesn't (this branch's own target query has no `LIMIT` anywhere, so it's
unaffected), and the exact pre-Phase-2 logic applies again when one
exists, so a prelimit source is only promoted when genuinely safe
regardless of the limit.

### Further gaps found while verifying the target query end-to-end

1. **`probe_work_rows` threading**: the filter-condition lowering path
   hardcoded a `nil` cost estimate regardless of the promotion gate, so
   any promoted stage reachable only through a WHERE-clause filter (as
   opposed to a projected field) always fell back to the safe
   materialized-query-probe carrier. Threaded the real
   `projection_probe_work_rows` estimate through
   `build_join_scan_leaf_using_recipe` and its recipe-function chain.
2. **Dependency-catalog gap**: neither the keytable nor RecSet carrier
   received the caller's dependency-closure stage catalog, unlike the
   safe fallback — causing a "references unknown stage" panic the moment
   either carrier became reachable from a filter-condition call site.
3. **Boolean-shape recognition gap**: `stage_boolean_shaped?` couldn't
   recognize a bare `SELECT sa.allowed FROM ... LIMIT 1` (a plain `BOOL`
   column, no comparison/CASE wrapping it) as boolean-shaped. Fixed by
   resolving a bare column leaf's declared SQL type via schema lookup
   when structural recursion bottoms out.
4. **RecSet per-row `(context "session")` cost** (the deepest finding):
   once carrier reachability was fixed, the RecSet carrier's one-time
   build turned out to be memoized via `(context "session")` read/write
   embedded directly in the per-row generated code. `(context "session")`
   is backed by `gls` (goroutine-local storage — Go has no native
   equivalent), which locates the current goroutine by walking the call
   stack in batches of 16 frames. Measured directly: ~20µs from a
   shallow call, but **~30ms per call** from deep inside a multi-level
   correlated subquery's per-row scan callback — regardless of whether
   the underlying value is already built, since the guard check itself
   pays the same tax every row. Fixed by adding a compile-time
   lexical-prepare accumulator (`queryplan_record_lexical_prepare` /
   `queryplan_drain_lexical_prepares` / `with_queryplan_lexical_prepares`),
   using the same `(context "session")` mutable-accumulator idiom
   `planner_record_guard_condition` already establishes for guard
   conditions, but running entirely at compile time: the RecSet carrier
   registers its one-time build expression while the plan is being
   built, and the compiled plan wraps itself in a single outer `!begin`
   binding every accumulated lexical variable once, so per-row generated
   code resolves it via normal (cheap, O(1)) lexical scoping instead of
   ever touching `(context "session")`.

## Verification

- `acl-boolean-membership-scaling.yaml` (100k rows, no LIMIT): went from
  a 30s+ timeout past its `max_time: 1s` gate to **~0.3–0.5s**, with the
  correct result, and EXPLAIN confirms a bounded physical carrier
  (`recset_key_index`/`scan_recset`) is now actually chosen.
- `order-limit.yaml`: regression fixed, timing back to master's exact
  speed.
- Full 238-file suite (`make test`, `MEMCP_FAIL_FAST=0`): passes clean.

Also carries a separate, pre-existing correctness bug fix
(`lower_recset_scalar_first_probe_expr`'s `once`/`apply` misuse, found
while investigating this work but predating and unrelated to it) —
already submitted and open as #533.

## Test plan

- [x] `go build -o memcp .`
- [x] `python3 tools/lint_scm.py --path lib/queryplan.scm --check`
- [x] Full in-process unit test suite: 1231/1233 (2 pre-existing,
      unrelated failures)
- [x] `make test` (full 238-file suite, fail-fast disabled): all pass
- [x] Targeted timing verification on both the target query and the
      regression case, against a clean `origin/master` baseline binary

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -125,6 +125,83 @@ deduplicate identical conditions without mutable Scheme lists. */
 			(if chosen condition (list (quote not) condition)))
 		chosen)))
 
+/* A physical carrier's one-time build (e.g. lower_recset_scalar_first_probe_expr's
+recset_key_index) cannot afford to memoize itself via (context "session") read/write
+embedded in the generated per-row IR: (context "session") is backed by gls (goroutine-
+local storage -- Go has no native equivalent), which locates the current goroutine by
+walking the call stack in batches of 16 frames looking for a marker. From a shallow
+call this costs ~20us; from deep inside a multi-level correlated subquery's per-row
+scan callback (many nested interpreter frames from evaluating the compound expression
+tree on every row) it costs on the order of tens of milliseconds PER ROW -- measured
+directly, not estimated -- which defeats the entire purpose of a one-time build no
+matter how well-guarded the memoization logic is, since the guard check itself pays
+the same tax on every row regardless of whether the underlying value is already built.
+
+queryplan_record_lexical_prepare instead runs the accumulation AT COMPILE TIME: this
+function executes as plain Scheme code while build_queryplan is constructing the plan,
+not as part of the generated IR, so calling (context "session") here costs the same
+~20us regardless of how deep the *eventual* per-row runtime call stack will be. It
+records a (lexical-variable-name . one-time-build-expression) pair in a compile-scoped
+accumulator -- the same (context "session") mutable-accumulator idiom
+planner_record_guard_condition already establishes above for guard conditions, and for
+the identical reason: it substitutes for threading state through every functional
+planner return value. queryplan_drain_lexical_prepares, called once after the whole
+plan is built (see neumann_compile_pipeline/neumann_compile_ir_pipeline), wraps the
+compiled plan in a single outer `let` binding every accumulated lexical variable to its
+one-time build expression, so per-row generated code that references the variable
+resolves it via normal (cheap, O(1)) lexical scoping -- never touching (context
+"session") at all. */
+(define queryplan_record_lexical_prepare (lambda (lex_key build_expr)
+	(begin
+		(define collector (try
+			(lambda () ((context "session") "__memcp_queryplan_lexical_prepares"))
+			(lambda (_e) nil)))
+		(if (nil? collector)
+			false
+			(begin
+				(if (nil? (collector lex_key)) (collector lex_key build_expr) true)
+				true)))))
+
+(define queryplan_drain_lexical_prepares (lambda ()
+	(try
+		(lambda ()
+			(begin
+				(define collector ((context "session") "__memcp_queryplan_lexical_prepares"))
+				(if (nil? collector)
+					'()
+					(map (collector) (lambda (key) (list (symbol key) (collector key)))))))
+		(lambda (_e) '()))))
+
+(define with_queryplan_lexical_prepares (lambda (compile_fn)
+	(begin
+		(define session (try
+			(lambda () (context "session"))
+			(lambda (_e) nil)))
+		(if (nil? session)
+			/* No session installed at all (e.g. a direct build_queryplan_term
+			call from a test harness outside any request/session context):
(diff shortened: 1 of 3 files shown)


carli2:
Superseded by #535, which continues the work with explicit value-vs-truth semantics, exact dependency resolution, and per-join-node physical costing.

PR #555, opened 2026-08-21
planner: keep membership carrier choices node-local

carli2:
## What

Make each physical membership consumer use the strategy and RecSet expression chosen for that exact tree edge.

Previously, costing could choose `candidate_keyset`, but later lowering reconstructed the decision from enclosing block facts and emitted a base-table scan with `$recset_contains`. The choice was visible in `EXPLAIN PHYSICAL` but did not control the executable operator.

This change:

- returns `(strategy, RecSet expression)` from the membership cost decision;
- makes ordered and join-leaf consumers use that exact pair once;
- feeds selective projected RecSets directly into `scan`/`scan_order`, including across downstream join continuations;
- keeps broad `driver_order_membership_probe` choices on the established direct driver-probe path instead of materializing a candidate set;
- documents the current row-number capability boundary, where the operator still requires a base-table source.

No calibration constants or logical IR shapes are changed.

## Regression proof

The new integration case reproduces an ordered derived document shape with:

- a selective UNION membership query;
- projection from the searched relation to the driving relation;
- a correlated ACL-style `EXISTS` predicate;
- a downstream `LEFT JOIN`;
- `ORDER BY ... LIMIT`.

On `master`, physical costing reports `candidate_keyset` but emitted code still contains `$recset_contains`. On this branch, the same query reports and emits:

- `chosen candidate_keyset`;
- `chosen scan_order_recset_part`;
- `recset_scan_source`;
- `recset_project_join`;

and contains neither `$recset_contains` nor `source_is_base_table` for that edge.

The real captured query shows the same physical diff: the previous saved plan had `candidate_keyset` plus `chosen recset_contains_probe`/`source_is_base_table`; this branch has two candidate-keyset decisions, `chosen scan_order_recset_part`, two `recset_scan_source` occurrences, and no membership-filter fallback.

Observed end-to-end timings on the restarted data-copy server were 131.95 s for the first cold file-search load, 5.61 s warm, and 89.7 ms from query cache. A new zero-hit search term took 5.45 s warm and 3.9 ms cached. The remaining first-load/string-search cost is separate from this carrier-selection bug.

## Validation

- Scheme startup unit tests: 1239/1239
- `tests/planner/subqueries/in-subquery.yaml`: 70/70
- `tests/planner/subqueries/acl-boolean-membership-scaling.yaml`: 13/13
- `tests/planner/order-window/order-limit.yaml`: 47/47, repeated three times
- `tests/execution/operators/recset.yaml`: 34/34
- repartition concurrency suite: 143/143, repeated
- rebuild concurrency suite: 16/16, repeated three times after an unrelated shared-host process interruption
- Scheme lint, `git diff --check`, and build pass

The full local hook was attempted while other worktrees were running concurrent coverage/calibration jobs. Its planner suites passed; shared-host server restarts made unrelated storage concurrency cases flaky. CI is therefore the authoritative full-suite gate for this draft.

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -12831,11 +12831,10 @@ the auto-index chooses the concrete access path on both tables. */
 					(exists_recset_project_join_expr src stage)
 					nil))))))
 
-/* A membership requirement becomes physical only here. Returning nil keeps
-the marker in the filter, whose ordinary lowering emits the driver-side
-scan_exists/keytable path. Returning the expression emits the RecSet carrier.
-That makes the recorded and forced choice identical to the executable plan. */
-(define recset_project_join_expr_for_membership_using (lambda (src membership consumer ordered_filter driver_rows_override)
+/* Cost one physical tree edge once and return (strategy RecSet-expression).
+Consumers decide whether that RecSet is their scan carrier or a membership
+filter; they must not reconstruct the choice from enclosing block facts. */
+(define recset_project_join_plan_for_membership_using (lambda (src membership consumer driver_rows_override)
 	(begin
 		(define stage (nth membership 0))
 		(define facts (gs_facts stage))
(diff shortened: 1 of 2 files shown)


carli2:
Superseded by #560, which integrates this node-local carrier work with the selective-filter lowering and the latest planner-statistics changes.

PR #556, opened 2026-08-21
perf: use selective text filters as RecSet carriers

carli2:
## Summary

- lower highly selective, source-local text predicates on the physical join driver into an exact `scan_recset` carrier
- keep ACL and join predicates residual so they run only for matching record IDs, while broad filters and single-source scans retain the fused scan path
- expose the selectivity decision in `EXPLAIN PHYSICAL` and preserve session state/query generation in fanout workers for cold cache initialization

## Selection guard

The carrier is chosen only for a multi-source physical driver with at least 1024 input rows, no unique-point binding or row-number stage, and an estimated density of at most 1/8. The normalized logical IR remains free of RecSets and scan artifacts.

## Measurement

On the production-copy snapshot, the cold exact query completed in 44.58 s with the new path. The unchanged baseline did not complete within 240 s on the same snapshot (at least 5.4x). The physical estimate selected 134 of 855,496 document rows. A manually forced RecSet rewrite on the live workload previously measured 6.82-8.87 s versus 87.16 s for the original shape.

## Verification

- `make test` (exit 0: commit allowed; known noncritical master cases remain)
- `python3 run_sql_tests.py tests/execution/operators/range-scan-coverage.yaml` (105/105)
- `python3 run_sql_tests.py tests/planner/subqueries/acl-boolean-membership-scaling.yaml` (13/13)
- focused fanout session/query context Go tests
- cold production-copy query returns HTTP 200 without cache-initialization panic

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -7121,6 +7121,12 @@ source catalog. join_plan remains the single owner of physical join order. */
 			(expr_contains_broad_text_match? head))
 		_ false)))
 
+(define expr_contains_text_match? (lambda (expr)
+	(match expr
+		((symbol strlike) _value _pattern _collation) true
+		((quote strlike) _value _pattern _collation) true
+		_ false)))
+
 (define stage_input_contains_broad_membership_filter? (lambda (input)
 	(if (union_block? input)
 		(reduce (union_branches input) (lambda (found branch)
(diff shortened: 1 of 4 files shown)


carli2:
Superseded by #560, which integrates this selective-filter RecSet work with node-local carrier decisions and the latest planner-statistics changes.

PR #557, opened 2026-08-21
Fix physical membership calibration and estimates

carli2:
## Summary

- make index-restricted selectivity estimates explicit lower bounds instead of extrapolating them as table samples
- carry local cardinality, decision-site, filter/map-column, expression, and broad-text byte work into the final membership decision
- collect average string widths during the existing REBUILD scan and use them without generating alternative plans
- isolate CALIBRATE variants, validate decision/operator coverage, race dangerous alternatives in parallel, and cancel the loser after winner runtime + 50%
- teach costgen to report current vs updated error, fit tagged YAML workloads, and rewrite the generated planner constants
- add million-row driver, COUNT, ordered LIMIT, wide-expression, column-width, and holdout calibration shapes

## Calibration evidence

`make costgen` on this host:

- training median absolute error: **10.1% -> 6.8%**
- training decision ranking: **12/12** before and after fitting
- wide-text candidate estimate error: **-19.1% -> -4.6%**
- million-row COUNT candidate estimate: **-13.5% -> -0.0%**

The generated constants in `lib/queryplan.scm` now include a 4 ns/input-byte broad-text term and a recalibrated 215 ns/driver-input-row RecSet aggregate term. Exact nanoseconds remain secondary to preserving the measured cost inequality; the report prints both errors and plan rankings.

Combined `EXPLAIN PHYSICAL CALIBRATE` output is marked `shared_sequential_diagnostic` and `fit_eligible=false`; costgen uses isolated `CALIBRATE VARIANT` requests. Timed-out alternatives are reported as lower bounds, not fitted as exact durations.

## Validation

- `make costgen`
- `go test ./tools/costgen`
- `python3 run_sql_tests.py tests/execution/operators/range-scan-coverage.yaml` (99/99)
- `python3 run_sql_tests.py tests/planner/subqueries/in-subquery.yaml` (69/69)
- `make test` (all CI-active YAML suites; Scheme coverage 63,535/63,535 nodes)

### DIFF
--- a/lib/queryplan.scm
+++ b/lib/queryplan.scm
@@ -3955,14 +3955,14 @@ carrier selection remains a later, per-stage cost decision. */
 				(and (query_block? subquery)
 					(and (not (empty_list? (qb_sources subquery)))
 						(and (equal? (count (qb_fields subquery)) 2)
-						(and (equal? (qb_limit subquery) 1)
-							(empty_list? (extract_aggregates (query_block_first_expr subquery))))))))))))
+							(and (equal? (qb_limit subquery) 1)
+								(empty_list? (extract_aggregates (query_block_first_expr subquery))))))))))))
 
 (define btw2025_scalar_bundle_key (lambda (marker)
 	(concat "scalar-bundle:"
 		(serialize (list
-		(dependent_subquery_scope_backbone (quote scalar) (dep_subquery_query marker))
-		(dep_subquery_outer_sources marker))))))
+			(dependent_subquery_scope_backbone (quote scalar) (dep_subquery_query marker))
+			(dep_subquery_outer_sources marker))))))
 
 (define btw2025_scalar_bundle_output_budget 16)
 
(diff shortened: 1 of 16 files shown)


carli2:
Hot-path optimization added in `92aca15b6`:

- `CountEstimate()` is now a single atomic O(1) load and never calls `GetRead`, takes a shard lock, or lazy-loads storage.
- REBUILD publishes an exact-at-that-point table estimate; INSERT/DELETE adjust it once per batch. Staleness is explicitly preferred over blocking compilation.
- The engine publishes one immutable planner snapshot per table. A query-planning session caches it once per relation and resolves columns through a `FastDict`, instead of repeated `SHOW` scans and linear column searches.
- Correctness-sensitive `table_empty?` uses the separate locked `CountExact()` path, so approximate planner state cannot change SQL results.
- Unknown pre-REBUILD NDV still uses the current atomic row count as `fallback_row_count`, without rebuilding the column catalog on every DML statement.

Measured with 256 columns (`BenchmarkPlannerStatisticsLookup`, 5 runs):

- published hash lookup: 33.31–33.81 ns/op, 0 allocs
- former linear SHOW lookup: 2.889–3.063 µs/op, 0 allocs
- about 87–92x faster per column-stat lookup

Validation: full `make test` passed, targeted race tests passed, range-scan coverage 99/99 passed.

PR #568, opened 2026-08-22
fix: localize Proc optimizer metadata and scalar empty-domain semantics

carli2:
## Summary

- attach complete inferred Scheme return metadata to the concrete procedure identity instead of mutating a shared environment map
- keep optimizer metadata behind a Proc-local pointer so CloseProcedure and JIT value copies share identity without copying future locks or specialization caches
- preserve nested TypeInfo details such as fixed-list keys, element types, function parameters, return types, and side-effect facts
- keep Proc at 64 bytes on 64-bit targets by storing its bounded numbered-variable count as int32
- trust the finalized Proc.NumVars metadata when adapting optimized map/filter/reduce callbacks, avoiding a full callback-body traversal for every adapter
- account for Proc optimizer metadata and retained type descriptors in ComputeSize
- prevent RecSet lowering from crossing COALESCE(..., TRUE), where an empty correlated scalar domain must remain distinguishable from a matching false row
- remove the unused global TypeDescriptor.Variants placeholder

## Design

This PR establishes the ownership boundary needed for later parameter-specialized procedure variants without implementing the variant compiler prematurely.

Each evaluated lambda receives a fresh ProcOptimizerMeta identity. Copies of that concrete Proc share the pointer. Future variants should publish immutable snapshots for lock-free reads and coordinate only an identical Proc plus specialization-key miss; unrelated procedures and keys must compile independently.

Optimized lambdas already finalize NumVars after introducing numbered temporaries. OptimizeProcToSerialFunction therefore consumes that metadata directly instead of recursively rediscovering the highest numbered slot whenever queryplan Scheme code passes a callback to map, filter, reduce, or another list primitive. Direct Proc constructors remain responsible for setting NumVars when their bodies contain NthLocalVar values.

## Compile-time A/B evidence

Compared `master` at `4e3e86efa` with this branch at `33c42c948` using separate binaries and MemCP processes on the same host. For every workload, both processes received identical schema setup and identical `EXPLAIN COMPILE` SQL. After two warmups per process, 25 measured runs per revision were interleaved in randomized order. Values below are medians; the parenthesized value is the median absolute deviation (MAD).

| Complex query workload | `compile_total_ns` master | `compile_total_ns` PR | Delta | `planner_total_ns` master | `planner_total_ns` PR | Delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Selective nested UNION/EXISTS | 375.179 ms (7.157) | 353.894 ms (4.556) | **-5.67%** | 341.770 ms (6.629) | 324.003 ms (5.557) | **-5.20%** |
| 48 distinct scalar probes | 539.159 ms (6.128) | 513.787 ms (4.658) | **-4.71%** | 497.762 ms (4.442) | 473.501 ms (3.835) | **-4.87%** |
| 256 dependent metrics | 630.594 ms (12.622) | 567.905 ms (13.647) | **-9.94%** | 206.278 ms (6.976) | 189.147 ms (8.993) | **-8.31%** |

The external HTTP wall-time medians moved in the same direction: -5.55%, -4.67%, and -9.46%, respectively. CPU profiles before the change attributed about 2 seconds of a 58-second sample window to repeated `requiredNumberedSlots` callback-body walks; the final A/B result confirms that removing those walks improves complete complex-query compilation rather than only a microbenchmark.

## Validation

- complete normal-parallel pre-commit suite: pass
- `go test ./scm -count=1`
- `go test -race ./scm -run TestConcurrentSchemeReturnInferenceDoesNotMutateEnvironment -count=1`
- scalar subselect patterns: 25/25
- compound correlated boolean/RecSet suite: 20/20
- selective membership scaling suite: 13/13
- `git diff --check`

The structured metadata test verifies per-position number/string return types, not only the top-level kind and length. The concurrency test evaluates every optimized definition in an isolated environment and verifies both Proc metadata and the expected `append_mut` consumer rewrite.

`tools/lint_scm.py --check` still reports only the repository baseline negative-depth warnings; none of the reported lines were changed by this update.

### DIFF
--- a/lib/queryplan-optimize.scm
+++ b/lib/queryplan-optimize.scm
@@ -3534,19 +3534,22 @@ the logical lookup still carries an alias which no longer exists. */
 									(join_optimizer_facts_without_aliases
 										(qb_facts physical_block) (list (source_alias candidate)))))))))))))
 /* The other probe shape a group-stage-output alias can appear as: a plain or
-COALESCE-wrapped bare boolean passthrough (no ">0" count-encoding), the shape
-`COALESCE((SELECT bool_expr ...) ...)` compiles into for a scalar_single
-boolean probe, matched against the specific logical output alias. */
+COALESCE-false-wrapped bare boolean passthrough (no ">0" count-encoding), the
+shape `COALESCE((SELECT bool_expr ...), FALSE)` compiles into for a
+scalar_single boolean probe, matched against the specific logical output
+alias. A TRUE fallback is deliberately excluded: positive membership cannot
+distinguish a matching FALSE row from the empty scalar domain, while SQL
+requires COALESCE(NULL, TRUE) to preserve the latter. */
 (define stage_output_boolean_probe_term? (lambda (alias term)
 	(match term
 		((symbol coalesceNil) ((symbol get_column) tblvar _tbl_ignorecase _col _col_ignorecase) _default)
-		(and (equal? tblvar alias) (or (equal?? _default false) (equal?? _default true)))
+		(and (equal? tblvar alias) (equal?? _default false))
 		((symbol coalesceNil) ((quote get_column) tblvar _tbl_ignorecase _col _col_ignorecase) _default)
-		(and (equal? tblvar alias) (or (equal?? _default false) (equal?? _default true)))
+		(and (equal? tblvar alias) (equal?? _default false))
 		((quote coalesceNil) ((symbol get_column) tblvar _tbl_ignorecase _col _col_ignorecase) _default)
-		(and (equal? tblvar alias) (or (equal?? _default false) (equal?? _default true)))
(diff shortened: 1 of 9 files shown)


carli2:
Follow-up after the fresh ERPL run: fixed transitive aggregate-column renames across nested decorrelated stage outputs. Boolean folding can rename a child carrier column; the optimizer now propagates that positional interface rename through the complete stage DAG so producers and consumers stay synchronized. Added a YAML regression that fails on the previous revision with `stale stage-output column` and passes after the fix. The exact ERPL DMS query with session specialization now returns the expected document row instead of `Column does not exist`. Full hook-equivalent `make test` passes; the selective membership runtime guard was widened from an unstable 1.00 s to 1.25 s after two 7–30 ms-only threshold misses with correct results and plan assertions.

PR #572, opened 2026-08-22
perf: compact bigram postings and intersections

carli2:
## What

Follow-up to #571 that keeps the public `CompressedRecSet.AndMut` contract but replaces its expensive resident and query-time implementation:

- choose per posting between a packed `StorageInt` RecID list, packed `StorageInt` ranges, and a raw bitmap;
- remove rANS from postings after profiling showed that its small resident-size advantage did not repay its decode/build cost;
- intersect every source representation directly and in place with bitmap, positive-list, and range RecSet destinations (3×3), without normalizing through a copied bitmap;
- decode packed integer batches through new unboxed `StorageInt.GetValuesUInt32Range` / `GetValuesUInt32Multi` APIs with stride support;
- build postings sparsely and promote to a temporary bitmap only at list/bitmap break-even;
- keep the query-local intersection as one disposable bitmap for fast repeated `AndMut` and row membership;
- replace the resident bigram hashmap with sorted compact key/set/count arrays; the build-only ASCII lookup table and Unicode map are released before final compression;
- scan LIKE patterns directly as runes without allocating an intermediate bigram array.

The shard-local index still covers main rows only. Delta rows pass through the row matcher and remain checked by the existing residual `STRLIKE` predicate.

## Memory accounting

`bigramIndex.ComputeSize` includes the index struct, the capacities of all three sorted-table backing arrays, every concrete compressed posting struct, and all bitmap/`StorageInt` backing capacities. There is no resident hashmap. Tests independently reconstruct the owned size for bitmap, positive-list, and range postings and compare it to the stored accounting value.

This is more exact than older storage accounting such as `StorageSparse`/`StorageEnum`, which still uses fixed management estimates and often counts lengths rather than capacities. As elsewhere in storage accounting, Go allocator size-class rounding is not included because it is not observable from an allocation.

## SQL A/B/C benchmark

All timings and allocations below come from real SQL sent through MemCP's HTTP SQL compiler/executor, not hand-built storage plans. The same persisted fixture and query sequence was used for:

- A: pre-#571 master (`04af4e092`)
- B: merged #571 (`635a9832d`)
- C: this PR (`8a69be655`)

Fixture: 1,000,000 rows in one deliberately fixed shard; compressed text column exactly 40,765,703 B (38.88 MiB); 100 controlled matches each for `fox`, `needle`, and `casinoword`. The runner discovers and sums every shard through table metadata instead of assuming a shard count.

### End-to-end SQL latency

| SQL pattern | A: old master | B: #571 | C: this PR | C vs A | C vs B |
|---|---:|---:|---:|---:|---:|
| hit, 3 chars (`%fox%`) | 89.42 ms | 15.43 ms | 12.15 ms | 7.4× faster | 1.27× faster |
| hit, 6 chars (`%needle%`) | 83.92 ms | 14.64 ms | 11.49 ms | 7.3× faster | 1.27× faster |
| hit, 10 chars (`%casinoword%`) | 86.76 ms | 17.44 ms | 11.36 ms | 7.6× faster | 1.53× faster |
| miss, 2 chars (`%xy%`) | 85.28 ms | 13.47 ms | 10.76 ms | 7.9× faster | 1.25× faster |
| miss, 3 chars (`%zzz%`) | 85.79 ms | 12.20 ms | 11.14 ms | 7.7× faster | 1.10× faster |
| miss, 6 chars (`%zzzzzz%`) | 87.31 ms | 12.23 ms | 11.04 ms | 7.9× faster | 1.11× faster |
| miss, 10 chars (`%zzzzzzzzzz%`) | 89.41 ms | 16.49 ms | 12.01 ms | 7.4× faster | 1.37× faster |

### Cold build and allocations

The first row includes normal process/column cold-start work. “Activate index” is the following unseen-bigram SQL that actually deploys/builds the shard-local index. “Warm same SQL” repeats it.

| Phase | A time / mallocs / allocated | B time / mallocs / allocated | C time / mallocs / allocated |
|---|---:|---:|---:|
| cold process SQL | 503.95 ms / 9,054,583 / 227,914,872 B | 489.43 ms / 9,054,601 / 227,879,056 B | 496.89 ms / 9,054,624 / 227,931,400 B |
| activate index | 88.32 ms / 796 / 251,040 B | 1,083.47 ms / 12,477 / 436,865,280 B | 264.29 ms / 9,565 / 159,086,264 B |
| warm same SQL | 0.83 ms / 760 / 70,632 B | 14.07 ms / 813 / 116,448 B | 11.72 ms / 807 / 116,536 B |

Compared with #571, index activation is 4.10× faster, performs 23.3% fewer allocations, and allocates 63.6% fewer bytes.

Representative unseen 6-character miss allocations are 816 mallocs / 203,864 B on old master, 859 / 111,744 B on #571, and 848 / 74,336 B here. The row matcher itself remains at 0 allocations per input batch.

### Resident index size (`ComputeSize`)

| Build | LIKE index | Ratio to compressed text |
|---|---:|---:|
| A: old per-pattern cache state | 3,422 B | 0.01% |
| B: #571 | 4,080,329 B (3.89 MiB) | 10.01% |
| C: this PR | 5,775,312 B (5.51 MiB) | 14.17% |

C intentionally spends 1.62 MiB (41.5%) more than #571 on this repetitive fixture to remove rANS. A representation breakdown independently reconstructs the totals: #571 uses 110 rANS postings totaling 4,059,673 B; C uses 10 bitmaps totaling 1,250,320 B, 41 packed positive postings totaling 8,016 B, and 135 packed range postings totaling 4,511,424 B. Sorted-table overhead is only 5,208 B, so the increase is representation data rather than double-counted table or map memory.

The profile showed rANS size calculation and encoding dominating the previous cold build. The corrected benchmark quantifies the trade: +41.5% resident index memory for 4.10× faster construction, 63.6% fewer constructed bytes, and consistently faster arbitrary SQL terms. For the reported real dataset with about 60 MiB compressed text, this fixture's ratio projects to roughly 8.5 MiB of bigram index if its distribution is comparable.

## Profile

Before removal, rANS bitmap size calculation was 26.8% flat CPU and encoding another 8.9% flat (21.6% cumulative). In the final no-rANS SQL profile those functions disappear entirely; remaining index work is posting insertion, one-time bitmap statistics, and final `StorageInt` packing.

## Verification

- targeted compressed RecSet/StorageInt/index tests, including all representation pairings and randomized tail-word boundaries: pass
- `tests/storage/indexes/fulltext-like-index.yaml`: 85/85
- `tests/execution/operators/recset.yaml`: 34/34
- SQL coverage: `buildBigramIndex` 88.5%, posting `add`/`finish` 100%, sorted table and size methods 100%, candidate-pattern path 96%; direct compressed intersection branches additionally covered by storage tests
- `git diff --check`: pass

### DIFF
--- a/storage/compressed_recset.go
+++ b/storage/compressed_recset.go
@@ -27,7 +27,195 @@ type CompressedRecSet interface {
 }
 
 func (s *recSetShard) AndMut(dst *recSetShard) {
-	intersectRecSetShardMut(dst, s)
+	if dst == nil || dst.count == 0 {
+		return
+	}
+	if s == nil || s.count == 0 {
+		dst.kind = recSetRanges
+		dst.data = nil
+		dst.used = 0
+		dst.count = 0
+		return
+	}
+	if dst.universe != s.universe {
+		panic("recset mutable intersection: universe mismatch")
+	}
+	switch s.kind {
+	case recSetBitmap:
+		(&compressedBitmap{universe: s.universe, words: s.data}).AndMut(dst)
+	case recSetPositive:
+		positiveValuesAndMut(dst, s.listedValues())
+	case recSetRanges:
+		rangeValuesAndMut(dst, s.listedRanges(), s.count)
+	}
+}
+
+func positiveValuesAndMut(dst *recSetShard, source []uint32) {
+	if recSetShardIsFull(dst) {
+		dst.kind = recSetPositive
+		dst.data = append([]uint32(nil), source...)
+		dst.used = uint32(len(source))
+		dst.count = int64(len(source))
+		return
+	}
+	if dst.kind == recSetRanges {
+		maximumMatches := min(uint64(len(source)), uint64(dst.count))
+		if maximumMatches*4 < uint64((dst.universe+31)/32*4) {
+			ranges := dst.listedRanges()
+			result := make([]uint32, 0, maximumMatches)
+			rangePosition := 0
+			for _, candidate := range source {
+				for rangePosition < len(ranges) && ranges[rangePosition]+ranges[rangePosition+1] <= candidate {
+					rangePosition += 2
+				}
+				if rangePosition == len(ranges) {
+					break
+				}
+				if candidate >= ranges[rangePosition] {
+					result = append(result, candidate)
+				}
+			}
+			dst.kind = recSetPositive
+			dst.data = result
+			dst.used = uint32(len(result))
+			dst.count = int64(len(result))
+			return
+		}
+		ranges := dst.listedRanges()
+		result := make([]uint32, (dst.universe+31)/32)
+		rangePosition := 0
+		for _, candidate := range source {
+			for rangePosition < len(ranges) && ranges[rangePosition]+ranges[rangePosition+1] <= candidate {
+				rangePosition += 2
+			}
+			if rangePosition == len(ranges) {
+				break
+			}
+			if candidate >= ranges[rangePosition] {
+				result[candidate>>5] |= uint32(1) << (candidate & 31)
+			}
+		}
+		dst.kind = recSetBitmap
+		dst.data = result
+		dst.used = 0
+		dst.count = int64(countBitmap(result))
+		return
+	}
+	if dst.kind == recSetBitmap {
+		cursor := uint32(0)
+		for _, candidate := range source {
+			clearRecSetBitmapRange(dst.data, cursor, candidate)
+			cursor = candidate + 1
+		}
+		clearRecSetBitmapRange(dst.data, cursor, dst.universe)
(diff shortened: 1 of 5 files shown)


carli2:
Benchmark correction: the first PR text summed a hard-coded 17 shards, while that fixture had been accidentally repartitioned into 9,935 tiny shards. I replaced the complete A/B/C section with a fresh one-shard, 1,000,000-row SQL fixture and dynamic full-shard enumeration. The corrected resident totals are #571: 4,080,329 B and this PR: 5,775,312 B. The +41.5% is real and comes from removing 110 rANS postings, not from duplicated ComputeSize accounting; the PR now includes the representation breakdown and corrected latency/allocation numbers.

PR #576, opened 2026-08-23
fix: expose delta-only planner row estimates

carli2:
## Summary
- publish per-shard main and delta row counts through atomics
- let CountEstimate use the single-shard delta count before the first REBUILD
- keep the compiler path O(1), lock-free, and free of lazy storage loads
- maintain the counters across inserts, WAL replay, rebuild, repartition, and cache cleanup

## Validation
- go test -race ./storage -run '^(TestCountEstimateFallsBackToSingleShardDeltaWithoutLocking|TestCountEstimateDoesNotTreatPersistedRowsAsDeltaOnly|TestCountEstimateIsLockFreeAndPersisted)$' -count=20
- make test

The fallback is intentionally approximate: it reports the published delta length and does not inspect deletion state. A non-zero table-wide estimate still wins, and main-storage presence disables the delta-only fallback.

### DIFF
--- a/storage/partition.go
+++ b/storage/partition.go
@@ -872,6 +872,7 @@ func (t *table) repartitionDDLReadLocked(shardCandidates []shardDimension) {
 		}
 		// Install main_count — from this point, new inserts get correct recids
 		s.main_count = uint32(mainN)
+		s.plannerMainRows.Store(uint32(mainN))
 		deltaLen := len(s.inserts)
 		if deltaLen > 0 {
 			// Shift deletion bitmap: bits in [0, deltaLen) move to [mainN, mainN+deltaLen).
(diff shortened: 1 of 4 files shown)


carli2:
ERPL integration validation (full manual runner, MemCP PR branch, TracePrint enabled): setup and all 19 de_DE phase-1 browser tests passed. de_DE phase 2 did not pass: 6/11 passed, with four UI/load timeouts and one missing vacation-account result. Trace evidence shows the known derived-window alias planner failure (Column does not exist: userKonto.differenz); the vacation-account miss is consistent with the separately reproduced delta/index-prefix bug. Therefore this PR is not being claimed as a green full-ERPL validation. The failures reproduce the current master baseline and are outside this lock-free CountEstimate change.

PR #582, opened 2026-08-24
Scale planner timing gates to calibrated hosts

carli2:
## What
- apply the protected CPU calibration to planner-total and compile-phase timing gates
- keep plan-size and compile-metric limits unscaled so structural/O(N²) regressions remain hardware-independent
- extend the runner contract and protected-policy checks for both rules

## Verification
- `python3 tools/test_sql_runner_contract.py` (19/19)
- `python3 tools/test_sql_time_limit_guard.py` (19/19)
- `python3 tools/check_sql_time_limit.py`
- full hook-equivalent suite passed with the identical patch in #580

This is an explicitly approved administrative update of protected policy code. Its trusted-base self-protection checks are expected to reject the policy-file change; workflows remain enabled and the merged policy becomes the sharp trusted baseline immediately.

### DIFF
--- a/run_sql_tests.py
+++ b/run_sql_tests.py
@@ -307,6 +307,11 @@ def scaled_wall_clock_limit_ms(seconds: float, calibration: Dict[str, Any]) -> f
     return seconds * 1000.0 * calibration["scale"]
 
 
+def scaled_compile_time_limit_ms(reference_ms: float, calibration: Dict[str, Any]) -> float:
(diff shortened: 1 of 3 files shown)


carli2:
Superseded by the explicitly approved administrative sequence on #580: rebase, verify/fix CI, temporarily bypass the two policy guards for that merge, then restore them immediately after merge.

PR #600, opened 2026-08-24
Bound concurrent cold query compilation

carli2:
## Summary

- add a shared, cancellation-aware admission limit for cold SQL and PostgreSQL query-plan cache producers
- weight generated queries by source size so formula-tree-heavy plans reserve more compile capacity
- let light producers use spare capacity instead of waiting behind a queued full-capacity producer
- leave cache hits, query execution, and unrelated cache maps unrestricted
- add deterministic Go and YAML regressions for overlap, weights, and head-of-line blocking

## 12-worker cold-workload A/B

| | Before | After weighted admission |
|---|---:|---:|
| Phase workflows passed | 7/19 | 18/19 |
| Slowest SELECT | 46.05 s | 5.97 s |
| Second slowest SELECT | 43.21 s | 5.26 s |

Further traces found that FIFO weighted admission could make trivial session and update statements wait about 3.3 s behind one queued wide compile even while a slot was free. The non-FIFO capacity controller removes that head-of-line stall. Medium generated statements of 42–46 KiB took about 2.4–2.5 s alone but 4.7–9.1 s when admitted in pairs, which motivates the adjusted 16/32 KiB weight boundaries.

The latest cold parallel run reached 18/19. The remaining browser failure did not execute the expected write transaction; the same workflow passed in a 56.2 s isolated warm run. It therefore remains evidence of cold-load responsiveness, not evidence of committed rows being lost.

## Locking and cancellation

Admission happens in the cache producer goroutine after the cache-map mutex is released. It acquires no table or shard locks. Waiters only hold the admission mutex for an O(1) capacity check. Capacity changes wake blocked producers, and canceled compile contexts leave capacity untouched.

## Validation

- fail-first on FIFO admission: a light producer was blocked behind a queued full-weight producer despite spare capacity
- `go test ./storage -run TestCompileCache -count=20`
- `python3 run_sql_tests.py tests/execution/concurrency/group-cache-initialization.yaml`
- `make test` (passed after final changes)
- repeated 12-worker browser workloads against fresh databases
- `python3 tools/lint_scm.py --check`
- `git diff --check`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ require (
 	github.com/lib/pq v1.11.2
 	github.com/pierrec/lz4/v4 v4.1.26
 	github.com/ulikunitz/xz v0.5.15
-	golang.org/x/sync v0.19.0
 	golang.org/x/text v0.21.0
 	golang.org/x/tools v0.42.0
 	gopkg.in/yaml.v3 v3.0.1
@@ -51,6 +50,7 @@ require (
 	github.com/shopspring/decimal v1.4.0 // indirect
 	golang.org/x/exp v0.0.0-20241009180824-f66d83c29e7c // indirect
 	golang.org/x/mod v0.33.0 // indirect
+	golang.org/x/sync v0.19.0 // indirect
 	golang.org/x/sys v0.41.0 // indirect
 )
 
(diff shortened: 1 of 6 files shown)


carli2:
Closing this implementation after rebasing and retesting. The targeted restart suite passes, but the full MemCP suite still trips a planner-scaling gate and the fresh 12-worker application workload regressed to 10/19. SQL source length plus a global weighted admission budget is not a robust proxy for real compile work. A replacement should separately measure admission wait and compile phases and use structural compile metrics rather than retaining this heuristic.

PR #601, opened 2026-08-24
Cancel read-only scans between batches

carli2:
## Summary
- propagate the active query context into shard scans
- stop read-only scans at index-batch boundaries after cancellation
- preserve shard-atomic execution for mutation scans
- add a regression test for cancelling CPU-heavy work inside one shard

## Locking and overhead
- one existing session-state lookup per scan obtains the query context
- batch checks read `context.Done()` without locks or allocations
- no checks are added to row loops
- no storage locks are added or upgraded

## Validation
- regression on `origin/master`: background scan still active after the 8 s synchronization budget; suite/process drain took about 25 s
- patched regression suite: 31/31 passed in about 6.7 s in the direct A/B run
- `go test ./scm` passed
- targeted persistence rollback test passed repeatedly
- full `make test` passed

The full `go test ./storage` command also exposed an unrelated pre-existing unsafe persistence/mmap fault in a restart test; the exact test passes repeatedly in isolation and the complete SQL persistence/crash-recovery suite passed.

### DIFF
--- a/scm/processlist.go
+++ b/scm/processlist.go
@@ -225,9 +225,8 @@ func (s *SessionState) ClearCancel(seq uint64) {
 
 // IsKilled returns true if this session has been killed.
 //
-// Storage execution contract: callers may check cancellation while scheduling
-// shard jobs, but never after entering a shard. Shard execution is atomic and
-// must not contain cancellation checks in index, batch, or row loops.
+// Storage execution contract: mutation shards remain atomic. Read-only scans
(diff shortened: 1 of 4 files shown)


carli2:
Closing this approach. Cancellation within a shard adds checks to the hot scan path and weakens shard-atomic execution. Master already checks cancellation at shard scheduling boundaries; after removing the per-batch checks this branch has no useful change left.

PR #602, opened 2026-08-25
Preserve aggregate extensions on shared logical stages

carli2:
## Problem
A grouped window-query can discover the same logical carrier through multiple consumers. A later variant may extend the carrier with additional aggregates, but first-wins stage deduplication discarded those columns. Physical readers then referenced aggregate columns which no retained prepare stage created.

Blindly merging every same-ID group stage is also incorrect: independently shaped scalar carriers may share a canonical ID while retaining different domain and output contracts.

## Fix
- Mark window aggregate carriers which explicitly support aggregate extension merging.
- Merge only marked same-ID stages with identical carrier backbones and a true aggregate-superset relationship.
- Preserve the first carrier domain and output contract while retaining the aggregate extension.
- Keep unrelated same-ID scalar carrier variants independent.
- Add regressions for both the missing-extension failure and the unsafe-merge counterexample.

## Validation
- Full local make test passed.
- The grouped-window reproduction builds and executes with every referenced aggregate present.
- The repeated scalar-carrier reproduction executes without an unknown aggregate-column compiler failure.

### DIFF
--- a/lib/queryplan-logical.scm
+++ b/lib/queryplan-logical.scm
@@ -309,31 +309,47 @@ catalog for every comparison. The binding catalog is compile-local as well. */
 	(reduce (merge (coalesceNil parts '())) (lambda (items item)
 		(logical_append_unique items item)) '())))
 
+(define logical_contains_all? (lambda (items required)
+	(reduce (coalesceNil required '()) (lambda (contains_all item)
+		(and contains_all (contains? (coalesceNil items '()) item))) true)))
+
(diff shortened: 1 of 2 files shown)


carli2:
Updated after end-to-end A/B validation: the merge is now capability-scoped to explicitly marked window aggregate carriers, requires the same carrier backbone and an aggregate-superset extension, and leaves unrelated scalar carriers untouched. Both regression directions pass, as does the complete local make test suite.

PR #623, opened 2026-08-26
Reduce SQL parser memoization overhead

carli2:
## Summary
- avoid memoizing terminal parser matches in the SQL root parser while retaining memoization for composite and named grammar rules
- preserve left-recursive grammar handling and cover it with a focused Go regression test
- add a generated large CASE/OR UPDATE integration shape to the performance test taxonomy

## Evidence
- exact 1.86 MB generated UPDATE, repeated cold A/B: 12.78 s -> 9.91 s (22.5% faster)
- full `make test` passed, including the new 4,584-WHEN / 763-OR regression
- Scheme formatter check passed

This changes parser allocation/GC overhead only; the SQL AST and planner phase boundaries remain unchanged.

### DIFF
--- a/lib/sql-parser.scm
+++ b/lib/sql-parser.scm
@@ -1941,7 +1941,7 @@ arithmetic; leave expressions containing columns or functions untouched. */
 		(parser '((atom "ROLLBACK" true)) (list (quote tx_rollback) (list (quote context) "session")))
 		"" /* comment only command */
 	)))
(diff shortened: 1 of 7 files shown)


carli2:
This PR should wait for the separate lookup-cache PR to be completed and merged into master first. Once that cache foundation is in master, this branch can be rebased and the remaining planner work can be reviewed against the final cache API and semantics.

PR #636, opened 2026-08-27
Cache scalar lookup order keys for bounded scans

carli2:
## Summary

- lower a decorrelated scalar lookup used as the leading ORDER BY key to a canonical computed column on the driving table
- retain direct scalar probes for tiny relations using the existing generated per-probe cost and the shared 100 ms planning risk budget
- guard that choice with live table cardinality so data growth recompiles the cached plan
- feed the remaining boolean/ACL stages back through the existing costed RecSet-versus-probe lowering after the native order carrier is installed
- preserve LEFT JOIN NULL semantics and rely on the storage dependency graph for lookup-source and foreign-key invalidation

This is a physical carrier choice after logical decorrelation and join reordering. No physical artifact is introduced into the logical IR, and no fallback evaluation path is added.

## Query inventory

An exhaustive token-based inventory of the deployed application covered all central query and prepare call sites, all 34 DataView builders, notification cron paths, schema checks and migrations, sessions, DAV/file access, ID-list helpers, login, and dynamically assembled SQL. The additional UNION, EXISTS-over-UNION, hierarchical navigation, DDL, and notification shapes are represented by existing or targeted YAML suites. No further unsupported SQL family was found.

## Cost contract

Structural eligibility is separate from the operator choice. Session-dependent, internally ordered, indirect-key, and non-base lookup shapes remain on the ordinary path. For an eligible lookup, the generated scalar-probe coefficient estimates direct work. Below the shared risk budget the direct path wins; above it the reusable native order carrier is selected. The exact same inequality is emitted into the query-cache guard. No generated calibration constant or hand-written crossover was changed.

## A/B measurement

Synthetic deployed-shape fixture, 100,000 documents:

- current master, repeated SELECT: 20.21-20.28 s
- branch after the lazy canonical carrier exists: 118 ms on the first measured request, then 4.1-4.2 ms
- permanent 4,000-row suite, including setup and eight correctness/plan cases: about 1.9 s

The first access still pays the lazy reusable carrier construction. Subsequent notification runs and equivalent queries reuse it; source-row updates and local foreign-key changes invalidate affected rows.

## Correctness coverage

- exact ordered notification ACL and anti-join result
- INSERT SELECT workflow
- missing lookup row and NULL ordering
- lookup-source update invalidation
- local foreign-key update invalidation
- session-dependent lookup exclusion
- direct path for trivial relations
- EXPLAIN PHYSICAL carrier decision

## Local validation

- startup Scheme tests: 1270/1270
- notification scaling: 8/8
- notification workflow: 11/11
- nested correlated subqueries: 12/12
- ORDER/LIMIT: 48/48
- computed lookup invalidation: 19/19
- dbcheck DDL compatibility: 81/81
- Scheme lint, build, and diff checks pass

The full YAML suite is intentionally left to CI.

### DIFF
--- a/lib/queryplan-physical-expr.scm
+++ b/lib/queryplan-physical-expr.scm
@@ -1781,7 +1781,7 @@ would still have to project that value over the segment. */
 		(query_invariant_scalar_first_probe_key stage requested_col)
 		(list (quote lambda) '() expr))))
 
-(define lower_table_scalar_first_probe_expr (lambda (sources default_alias src stage value_expr keys lookup_keys order_exprs dirs offset_value partition_limit)
+(define lower_table_scalar_first_probe_expr (lambda (sources default_alias src stage value_expr keys lookup_keys order_exprs dirs offset_value partition_limit tx_expr)
 	(begin
 		(define condition (coalesceNil (qassoc_get (gs_facts stage) (quote condition) true) true))
 		(define condition_cols (extract_columns_for_alias src condition))
@@ -1791,7 +1791,7 @@ would still have to project that value over the segment. */
 		(define filtercols (merge_unique (list condition_cols key_cols order_cols)))
 		(define mapcols (merge_unique (list value_cols)))
 		(list (quote scan_order)
-			'(session "__memcp_tx")
+			tx_expr
 			(source_table_expr src)
 			(cons (quote list) filtercols)
 			(list (quote lambda)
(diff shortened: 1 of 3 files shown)


carli2:
CI follow-up: the first run exposed a performance regression in the existing unselective scalar ORDER/LIMIT suite. The carrier eligibility now additionally requires the scalar lookup keys to cover a complete unique key of the lookup source. Non-unique LIMIT 1 lookups therefore retain the established late-projection path (11/11 locally), while the primary-key notification lookup still selects the canonical carrier (8/8 locally). This is a structural operator precondition, not a query-specific exception.

PR #650, opened 2026-08-29
Optimize costed grouped hierarchy joins

carli2:
## What

- add a non-CI performance fixture for the MySQL forum-hierarchy workload from https://www.reddit.com/r/mysql/comments/1l0ceun/having_trouble_understanding_the_problem_point_in/ at 10,000 forums and 33,669 threads/posts, using the default shard size and rebuild-driven automatic partitioning
- preserve aggregate aliases when a grouped derived-table output feeds a later join
- retain eligible grouped relations until physical lowering and compare keyed direct aggregation with the canonical group carrier through the physical cost model
- reduce all aggregates requested from one grouped relation in a single keyed scan
- recognize a constant unique lookup as the dominant join driver before the hierarchy scan
- keep `scan_join_order` behind an exact projected carrier when their fuzzy cost intervals overlap, while retaining it for clear winners
- calibrate grouped-relation startup/build coefficients and the separate `scan_join_order` startup coefficient through Costgen
- fix keyed-probe costing: every outer row performs a lookup regardless of join match selectivity, while local source predicates use the existing sampled cardinality estimator
- fix direct-group code generation so aggregate values are passed as lambda arguments rather than as one list argument

The logical planner remains free of physical carriers and RecSets. The direct-group alternative is selected during physical stage preparation and is visible/forceable through `EXPLAIN PHYSICAL CALIBRATE`.

## Adaptive warm-cache policy

The direct grouped path now follows the same pay-for-build idea as automatic indexes, entirely in SCM:

- if the canonical group table already exists, prepare any aggregate columns newly required by this query once in the query prelude and use the cache
- otherwise execute the direct keyed probes and accumulate their measured nanoseconds once per query in `system_statistic.group_cache_candidates`
- the candidate table is `ENGINE=sloppy` and contains only the canonical cache name and accumulated time without the cache
- cross the build threshold using the existing Costgen-calibrated group-relation startup/build cost, then start the canonical group-table preparation asynchronously with `setTimeout`
- delete the candidate after a successful build; clean up a partial cache and rethrow the original asynchronous error on failure
- keep Costgen-forced variants free of telemetry and cache reuse so calibration still measures the requested operator

The table-existence guards use a promise backed by a stack-resident two-cell list. They distinguish a missing table from an exception without allocating a FastDict or hiding execution failures. The direct probes accumulate timings in a query-local session, so the system-statistics upsert happens once per query rather than once per joined row.

## A/B measurements

Reference x86_64 host, persisted automatically partitioned fixture, median of five timed executions:

| Query / alternative | Before | After |
| --- | ---: | ---: |
| Original Reddit SQL | 4,077 ms | 8.9 ms |
| Correlated SQL rewrite | n/a | 14.8 ms |
| Child aggregate only | n/a | 7.8 ms |
| Thread aggregate only | n/a | 8.2 ms |

Latest adaptive-policy verification on the same fixture:

| Query | Time |
| --- | ---: |
| Child aggregate only | 8.6 ms |
| Thread aggregate only | 8.4 ms |
| Combined original query | 11.0 ms |

Forced whole-query physical variants on the same persisted fixture:

| Decision | Direct grouped probe | Group carrier |
| --- | ---: | ---: |
| child-count relation | 4.76 ms | 53.42 ms |
| thread aggregate relation | 4.47 ms | 1,685.92 ms |

Costgen after correcting probe cardinalities:

- direct-group median absolute estimation error: 71.6% -> 1.5%
- direct-group operator choices: 2/2 correct
- general training decision coverage: 18/18
- general holdout decision coverage: 7/7
- forced alternatives produced identical row counts and hashes

The fuzzy ordered-join guard was checked against both sides of its crossover: an exact projected carrier stays at 1.0-1.13 ms where `scan_join_order` needs 3.46-3.74 ms, while the tiny ordered join keeps `scan_join_order` at 0.22-0.25 ms versus 0.59-0.71 ms for the carrier.

## Validation

- `go test ./tools/costgen`
- Scheme bootstrap: 1270/1270
- group-stage corner cases: 39/39
- forum correctness fixture: 6/6
- forum performance fixture: 10/10
- `python3 tools/lint_scm.py` and `git diff --check`
- the complete Costgen fixture and full regression suite are left to CI for the pushed follow-up commit

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -49,17 +49,21 @@ jobs:
       - name: Measure base and candidate with the SQL test runner
         shell: bash
         run: |
-          runner="$GITHUB_WORKSPACE/base/run_sql_tests.py"
-          if ! grep -q 'PERF_AB_MODE' "$runner"; then
-            runner="$GITHUB_WORKSPACE/candidate/run_sql_tests.py"
-          fi
+          # The candidate runner defines the A/B protocol under review (including
+          # shared prepare/restart metadata); MEMCP_TEST_WORKTREE still supplies
+          # the exact binary and Scheme library for each measured side.
+          runner="$GITHUB_WORKSPACE/candidate/run_sql_tests.py"
           baseline="$GITHUB_WORKSPACE/perf-baseline.json"
           seed="$GITHUB_WORKSPACE/base/tests/performance/ci-workloads.json"
           if [ ! -f "$seed" ]; then
             seed="$GITHUB_WORKSPACE/candidate/tests/performance/ci-workloads.json"
           fi
 
-          (cd base && env \
+          # Discover YAMLs from the candidate for both sides. This makes newly
+          # added performance suites part of the A/B baseline without copying or
+          # running a separate fill workload; MEMCP_TEST_WORKTREE still selects
+          # the exact base binary and Scheme library.
+          (cd candidate && env \
             PERF_AB_MODE=record \
             PERF_BASELINE_FILE="$baseline" \
             PERF_BASELINE_SEED="$seed" \
--- a/lib/queryplan-logical.scm
+++ b/lib/queryplan-logical.scm
@@ -5220,7 +5220,9 @@ that actually owns the title consumes the reference. */
 								(begin
 									(define staged (make_grouped_derived_stage_source src alias inner))
 									(list
-										(merge (list (nth staged 0) tail_sources))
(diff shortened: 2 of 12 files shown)


carli2:
direct group scans might be faster in some corner cases, but group cache lookups might still ammortize over time. There must be a mechanism similar to the autoindex decider on group caches.

PR #654, opened 2026-08-29
Add protected base-vs-candidate performance gate

carli2:
## Summary
- add an independent Performance A/B workflow that measures the exact PR base and merge candidate on the same runner
- fail closed when any trusted performance case is missing, uses a different row workload, or exceeds its protected 10-30% regression allowance (20% default)
- add a trusted pull_request_target policy that prevents deleting or weakening performance tests and protects the benchmark workflow/harness
- keep performance assertions separate from make test and the existing unit/SQL CI job

## Validation
- make test
- python3 tools/test_perf_ab.py
- python3 tools/test_performance_policy.py
- python3 tools/check_performance_policy.py --base-root . --candidate-root .
- real A/B smoke test on tests/performance/group-lookup-performance.yaml against identical revisions (pass, +2.2% with +20% limit)


carli2:
Superseded: performance CI will be rebuilt from master around run_sql_tests.py only.

PR #655, opened 2026-08-29
ci: parallelize performance fixture work

carli2:
Follow-up to #654. Runs existing performance YAML suites concurrently for setup and other unmeasured work while an inter-process shared/exclusive gate gives warmups and timed repetitions uncontended access to MemCP. Caps each performance setup at 120 seconds and retains the existing RAM-pressure abort. Adds a 15-minute CI stress job using the existing million-row Reddit fixture and physical cost calibration suite; no synthetic fill suites are introduced.


carli2:
Superseded by a clean implementation based directly on run_sql_tests.py.

PR #666, opened 2026-08-30
perf: make cached join guards constant time

carli2:
## What

- replace repeated derived join-cost guards with process-unique immutable planner-statistics tokens for base tables
- retain exact cardinality guards for derived and synthetic planner nodes
- discard unused guard-runtime bindings instead of evaluating abandoned cost expressions on every cache hit
- bind only guarded cache-dispatch formulas and adaptive observation sessions to the execution session already owned by `sql_execute_formula`, with a compatibility fallback for direct `cached_parse` users
- preserve the fully optimized physical plan before constructing that guarded dispatch; exact-cache formulas remain untouched
- reuse non-capturing `begin` scopes in serial lambda execution; dynamic closure-producing forms keep the allocating compatibility path

## Why

The WordPress posts/postmeta ORDER BY/LIMIT plan was already the desired combined `scan_join_order_batched_probe` path. Its cached-plan guard nevertheless recomputed six selectivity/cardinality expressions on every execution. On current `master`, the guard alone costs about 0.70 ms while the useful operator plan costs about 0.26 ms.

Table statistics are published as immutable snapshots. A process-unique token therefore represents the complete dependency cheaply: any row-count, column-statistics, rebuild, or DDL snapshot publication changes the token and invalidates the cached variant. Shapes without a base-table snapshot keep their previous exact runtime guard.

Session binding is deliberately limited to the guarded cache dispatch. Post-optimization rewriting of exact-cache formulas caused a 50% indexed-JSON regression and changed computed/range-index behavior. Conversely, feeding the dynamic execution-session symbol into optimization too early regressed a complex guarded JOIN+LIKE hierarchy. The final phase boundary keeps normal optimizer/JIT specialization intact and only binds the polymorphic dispatch which needs cross-request session reuse.

## Manual A/B measurements

These measurements compare current `master` (`892242857`, including #665 and #667) with this branch after merging that master. Both servers used the same 8,214-post/75,000-postmeta fixture and the same query from `tests/performance/wordpress-postmeta-order-limit.yaml`. Caches were warmed before measurement.

| Measurement | Current master | This PR | Change |
|---|---:|---:|---:|
| Persistent HTTP, mean of three alternating 10,000-request `ab -k -c 1` runs | 1.348 ms/request | 0.583 ms/request | -0.765 ms / -56.8% |
| Extracted cached-plan guard, 1,000 in-process executions | 0.696 ms/query | 0.0306 ms/query | -0.666 ms / -95.6% |
| Extracted useful plan, 1,000 in-process executions | 0.264 ms/query | 0.273 ms/query | +0.009 ms / +3.4% |

The useful operator timing is effectively unchanged; the live gain comes from removing repeated planner-cost work from the cache-hit guard. For comparison, MariaDB measured 0.425 ms for this SQL shape. The live MemCP path is now about 1.37x MariaDB rather than roughly 3x, with an absolute gap of about 0.158 ms/request in this harness.

Regression needles measured during development:

- indexed selective JSON projection: 0.795 ms master, 0.801 ms final branch (effectively unchanged)
- aggregate-pushdown cold compile, 20 samples: 47.04 ms median master, 48.02 ms final branch (+2.1%)
- range-scan statistics: 113/113 after restoring the optimizer boundary
- computed index columns: 39/39 after restoring the optimizer boundary
- an eager lexical-callgraph preparation experiment measured 53–54 ms cold and was removed

## Focused verification

- `tests/sql/expressions/prepared-stmts.yaml`: 54/54
- `tests/planner/subqueries/decorrelated-join-reordering.yaml`: 8/8
- `tests/planner/order-window/order-limit.yaml`: 51/51
- `tests/performance/session-group-cache.yaml`: 37/37
- `tests/performance/wordpress-postmeta-order-limit.yaml`: 4/4
- `tests/performance/forum-hierarchy-aggregate-joins.yaml`: 9/9
- focused `go test ./scm ./storage` coverage for SerialProc, planner-statistics tokens, ownership interaction, and scan_join_order passes
- server startup Scheme tests: 1270/1270

No local full test run was performed, per repository optimization guidance.

`TestJITExpressionConsTailOfSingleElementList` still fails identically on the merged current `master`; this branch does not touch that matcher/JIT compilation path.


carli2:
Abandoning this approach. CI exposed two real planner/index regressions beyond the cascading performance-runner failures: `range-scan-coverage` observed candidateCount 10 instead of 200, and `computed-index-cols` lost the selective JSON_VALUE index (16 candidates instead of 4). A rerun also reproduced an unrelated planner workload regression (+29.3%). The process-unique statistics token is therefore too coarse for the existing cache invalidation and plan-selection contracts. Keeping the branch for forensic comparison; a future attempt needs narrower dependency tokens and full-suite validation before opening.

PR #678, opened 2026-08-30
optimizer: move rewrite hooks to declarations

carli2:
## Problem

#676 put a hard-coded producer whitelist and additional capture state into the central `optimizeList` begin path. That couples a generic def/use pass to concrete builtin names and makes the already central optimizer path larger. It also prevented the architecture from naturally gaining new operator patterns through declarations.

## Change

- fully revert the merged #676 producer whitelist and its specialized tests
- move `Optimize` from `TypeDescriptor` to `Declaration`
- move `OptimizeFirstArgTransfer` with it, because consumption is optimizer behavior rather than value type information
- dispatch declaration rewrites through `callDecl.Optimize`
- migrate all 44 existing optimizer hooks and all 15 first-argument ownership contracts in `scm` and `storage`
- add a direct regression test proving that a declaration-owned hook is dispatched

No extra AST traversal or subtree analysis is introduced. `TypeDescriptor` is again limited to type, ownership, escape, and return-flow facts; declarations own executable rewrite policy.

## A/B microbenchmark

Current `master` versus this PR, `BenchmarkOptimizeGeneratedConsChain32`, 15 samples per side, 1 second per sample:

| metric | master | PR | delta |
|---|---:|---:|---:|
| time | 4,750 ns/op | 4,532 ns/op | -4.58% |
| bytes | 3,920 B/op | 3,888 B/op | -0.82% |
| allocations | 46 allocs/op | 46 allocs/op | unchanged |

This benchmark repeatedly runs a declaration-specific optimizer hook. Moving function pointers out of copied type descriptors also removes 32 bytes per optimization.

## Cold SQL A/B

20 alternating cache-busted `EXPLAIN COMPILE` samples per side against identical fresh servers and a ten-table equality join:

| metric | master | PR | delta |
|---|---:|---:|---:|
| `compile_total`, mean | 69.284 ms | 67.953 ms | -1.92% |
| `compile_total`, median | 65.517 ms | 65.692 ms | +0.27% |
| optimizer, mean | 0.102 ms | 0.099 ms | -2.92% |
| optimizer, median | 0.093 ms | 0.087 ms | -5.86% |
| wall, mean | 71.464 ms | 70.191 ms | -1.78% |
| wall, median | 67.497 ms | 67.523 ms | +0.04% |

The cold compile result is neutral within run-to-run noise; there is no end-to-end regression signal.

## Validation

- focused declaration dispatch, optimizer, ownership, and storage optimizer tests: green
- complete Scheme startup suite: 1270/1270 green
- `git diff --check`: green
- full repository suite delegated to CI per maintainer direction

A broad local `go test ./scm ./storage` also encounters `TestJITExpressionConsTailOfSingleElementList` from freshly merged #674; the exact test fails identically on untouched `master` and this branch. It is intentionally not changed here because this PR is the optimizer architecture fix, not the JITGen workstream.


carli2:
Superseded by the squash merge of #679: #679 was stacked on this branch, so current master already contains the complete #678 content. A direct tree comparison shows that master differs from #678 head only by #679\x27s range-consumer additions in scm/list.go and scm/list_pipeline_optimizer_test.go; all declaration-hook and whitelist-removal changes from #678 are already byte-identical on master. The performance reruns therefore no longer gate an unmerged code delta.

PR #691, opened 2026-08-31
optimizer: share foldable scalar expressions

carli2:
## What

- add begin-local common-subexpression elimination for calls whose declaration is `Foldable`
- restrict sharing to immutable scalar parameter/return types, stable bindings, guaranteed execution paths, and non-shadowed declarations
- discover/hash/rewrite in one tree walk; only hash collisions/duplicates receive a structural equality check
- lower the common boolean regex form
  `(match value (regex "pattern" _) true _ false)`
  to a precompiled `MatchString` predicate without a match environment or capture slice

For `cached_parse`, the optimized body now computes `(toUpper parse_query)` once and feeds all three SQL classifiers from that local value. Its three capture-free boolean regex matches lower to precompiled predicates.

Fresh list/assoc results are deliberately not shared: their ownership identity remains observable by mutating consumers. Lazy branches are never selected as the first evaluation site, and calls are not lifted across an earlier argument.

## A/B measurements against `fe6127baa` (master)

Ryzen 9 7900X3D, Linux, `count=10`; same benchmark source on both revisions:

| benchmark | master | PR | delta |
|---|---:|---:|---:|
| repeated uppercase classifiers | 1,097 ns/op, 552 B/op, 11 allocs | 808.5 ns/op, 376 B/op, 7 allocs | **-26.3% time, -31.9% bytes, -36.4% allocs** |
| boolean regex match | 243.6 ns/op, 81 B/op, 3 allocs | 175.0 ns/op, 81 B/op, 2 allocs | **-28.2% time, -33.3% allocs** |

`benchstat` reports `p=0.000`, `n=10` for both time improvements.

Warm SQL cache path, 12 alternating samples of 100,000 calls to the real `cached_parse` with `SELECT 1`:

- master median: **13.44 us/call**
- PR median: **12.29 us/call**
- delta: **-8.5%**

End-to-end HTTP `SELECT 1` remains neutral within frontend/socket noise (CPU-pinned sequential runs: about 151-152 us median/request). Cold `EXPLAIN COMPILE` is likewise neutral (4.174 ms vs 4.168 ms median); parsing/planning dominates that case, as expected.

## Producer-lambda experiment

This PR does not claim a producer-lambda win. Moving a lambda binding closer to an ordinary call does not delay construction because Scheme evaluates call arguments eagerly. Safely sinking captured work through a callback additionally needs parameter metadata distinguishing at-most-once callbacks from repeated callbacks (`map`/`filter`/`reduce`). Adding an unused flag would not improve this path, so callback cardinality plus actual lazy materialization is left as a separate measured follow-up.

## Validation

- `go test ./scm -count=1`
- 1,270/1,270 Scheme startup tests in both A/B binaries
- focused correctness tests for lazy paths, argument order, shadowed callables, ownership-bearing results, regex captures, and regex type errors

The repository-wide SQL full test is left to CI as agreed for optimizer work.

### DIFF
--- a/scm/optimizer.go
+++ b/scm/optimizer.go
@@ -1868,6 +1868,12 @@ func optimizeList(v []Scmer, env *Env, ome *optimizerMetainfo, useResult bool) (
 				}
 			}
 		}
+		var cseBindings int
+		v, cseBindings = optimizeBeginFoldableCSE(v, bodyStart, ome)
+		if headSym == Symbol("begin_mut") && cseBindings > 0 {
+			reserve += cseBindings
+			v[1] = NewInt(int64(reserve))
+		}
 		usedVariables := make(map[Symbol]int)
 		variableContent := make(map[Symbol]Scmer)
 		bindings := make(map[Symbol]localBindingFacts)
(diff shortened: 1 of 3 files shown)


carli2:
CI follow-up: the first `performance-ab` run stopped on its single cold sample for `aggregate pushdown cold FK partition count` (60.4 ms base vs 119.1 ms candidate); all later “missing in candidate” reports were cascade failures after that abort. Local isolated A/B did not reproduce the regression (cold suite: 125 ms base vs 104 ms candidate), and 20 alternating `EXPLAIN COMPILE` samples were neutral at 39.69 ms vs 39.61 ms median. Re-running the unchanged commit passed `performance-ab` in full. All four required checks are now green.

PR #696, opened 2026-08-31
Test MemCP-to-MariaDB backup and restore in CI

carli2:
Adds a disposable recovery testbench under tools/ and runs it as a separate CI job against MariaDB 10.11.

The test starts MemCP, creates a synthetic application-shaped fixture, invokes an unmodified mysqldump, restores the resulting SQL into MariaDB and a fresh MemCP instance, and validates rows, binary BLOB bytes, empty-table schemas, unique indexes, AUTO_INCREMENT continuation, foreign-key cascade behavior, and trigger restoration.

No recovery test or documentation is placed under tests/; this external testbench remains under tools/.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,3 +1,18 @@
+# Copyright (C) 2026  Carl-Philip Haensch
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


carli2:
Rejected. The PR dosen't fix the mysqldump capabilities of memcp and does some quirky things.

PR #700, opened 2026-08-31
Support mysqldump recovery through the SQL compiler

carli2:
Replaces the rejected implementation from PR #699 with a clean single-parser design.

Implements mysqldump and mariadb-dump round-trip compatibility while keeping lib/sql-parser.scm as the sole SQL parser.

Architecture:

- the existing SQL compiler expands the final projection and reports result-field titles through a structured callback before execution
- the MySQL RowWriter consumes that compiled metadata, including for zero-row SELECT results
- all Go-side keyword, identifier, and projection parsing from the rejected revision has been removed
- the older Go-side isSelectQuery classification has also been removed

SQL and storage compatibility fixes:

- accept MariaDB-style 0x binary literals in the existing SQL parser
- preserve restored explicit AUTO_INCREMENT values without consuming generated IDs

The existing CI test job runs one recovery suite covering all four directions:

- MemCP dump -> MariaDB restore
- MemCP dump -> MemCP restore
- MariaDB dump -> MariaDB restore
- MariaDB dump -> MemCP restore

Every restore verifies table and column sets, row counts and aggregates, binary BLOB bytes, empty tables, unique constraints, AUTO_INCREMENT continuation, foreign-key cascade behavior, and restored trigger execution.

Targeted validation on the rebased final commit:

- build: passed
- focused MySQL metadata unit tests: passed
- INSERT edge cases: 31/31 passed
- AUTO_INCREMENT regressions: 64/64 passed
- recovery matrix: all four directions passed with complete integrity checks

The full repository suite is left to PR CI as requested.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,3 +1,18 @@
+# Copyright (C) 2026  Carl-Philip Haensch
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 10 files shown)


carli2:
rejected again for the same reason. don't introduce new parsers in to. We already have the parser infrastructure in scm/parser.go and the sql grammar in lib/. Please extend this code branch.

PR #710, opened 2026-09-01
Stabilize performance A/B sampling

carli2:
## Summary

- extend adaptive A/B measurements from at most 100 samples / 250 ms to 1000 samples / 1 s
- model shared-runner jitter as one fixed 50 ms budget for the complete measurement block, amortized across its samples instead of weakening the 20% per-query regression limit
- scale the three non-repeatable INSERT fixtures to 30k rows because repeating a mutation would change the measured workload

The round barriers, correctness checks, RAM guard, and 20% relative regression limit remain unchanged.

## Evidence

The same unchanged #707 candidate produced contradictory `Perf: GROUP BY` results on the same GitHub runner class:

- 1.377 ms -> 1.684 ms (+22.3%, failed)
- 1.390 ms -> 1.379 ms (-0.8%, passed in the complete diagnostic run)
- 1.466 ms -> 1.773 ms (+20.9%, failed)

Those runs were capped at 100 samples, totaling only 138-177 ms despite the configured 250 ms target. The new 1 s target and 1000-sample cap amortize the fixed CI scheduling/HTTP noise. A 50 ms block budget contributes 0.5 ms at 100 samples, but only 0.05 ms at 1000 samples; the relative 20% gate remains the dominant criterion for properly sampled cases.

Single-shot INSERT timing at 300k rows was measured locally as 5.65 s / 9.26 s / 2.99 s. The committed 30k fixture keeps these mutation cases long enough to avoid the former 8-28 ms one-shot measurements without adding that 31-second suite cost.

## Verification

- `python3 -m unittest tools.test_sql_runner_contract` (46 tests, green)
- both edited YAML suites parse successfully
- full SQL suite intentionally left to CI

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -67,8 +67,9 @@ jobs:
             PERF_AB_MODE=record \
             PERF_BASELINE_FILE="$baseline" \
             PERF_BASELINE_SEED="$seed" \
-            PERF_REPEAT=100 \
-            PERF_MIN_MEASURE_MS=250 \
+            PERF_REPEAT=1000 \
+            PERF_MIN_MEASURE_MS=1000 \
+            PERF_AB_JITTER_MS=50 \
             PERF_SETUP_MAX_TIME_SEC=120 \
             MEMCP_TEST_JOBS=4 \
             MEMCP_TEST_WORKTREE="$GITHUB_WORKSPACE/base" \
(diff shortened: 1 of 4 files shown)


carli2:
CI A/B validation is green (4m55s):

- `Perf: GROUP BY` now collected 705/711 samples and ~1.00s on each side: 1.419ms -> 1.408ms (-0.8%).
- The scaled mutation trials measured 637ms -> 711ms, 762ms -> 738ms, and 222ms -> 253ms.
- The deliberately cold single trial still showed the expected shared-runner jitter (59.9ms -> 88.9ms); its 29ms absolute delta stayed within the one-time 50ms block budget.

This demonstrates that the fixed block budget is amortized away for repeatable queries while stabilizing the genuinely non-repeatable cold measurement.

PR #711, opened 2026-09-01
Fix residual correlated presence probes

carli2:
## What

- keep residual-correlated `EXISTS` stages inside the enclosing scalar probe
- prevent a predicate that still reads an outer row from being prepared as a closed group cache
- add an anonymous integration regression covering RecSet deletion visibility, SQL `DELETE`, compound ACL shapes, persistent session rebinding, ID-only derived projections, and `INSERT ... SELECT` into an ID helper table

## Root cause

The scan family already applies transaction visibility and delete masks, including when its source is a precomputed RecSet. The observed ID-list failure happened earlier: an `EXISTS` predicate with an outer reference below `OR` retained that reference as a residual domain dependency. Scalar probe lowering nevertheless allowed its presence stage to be prepared independently as a group cache. That initializer had no enclosing domain row and produced false/empty ACL results, so the application selected no IDs to delete.

Residual presence probes have no correct closed-cache alternative. The lowerer now treats evaluation in the enclosing row scope as a correctness-dominance decision; ordinary closed direct-vs-carrier alternatives remain cost-selected.

## Verification

On unmodified master, the new suite reports 8/11 passing. The residual-domain ACL query and both unrestricted ID-projection executions fail. The low-level RecSet/delete checks pass, confirming that delete masks are not the defective layer.

On this branch:

```
python3 run_sql_tests.py tests/integration/regressions/deleted-row-carrier-visibility.yaml
11/11 passed
```

Also checked locally:

- Scheme startup unit tests: 1270/1270
- `tests/planner/subqueries/deep-correlation-membership.yaml` (existing noncritical expectations only)
- `tests/planner/subqueries/compound-boolean-recset-selection.yaml`: 23/23
- `tests/planner/subqueries/exists-subquery.yaml`: 35/35
- `tests/integration/regressions/nested-aggregate-carrier.yaml`: 1/1
- `tests/planner/joins/prejoin-update-delete.yaml`: 10/10

The full SQL suite is left to CI.

### DIFF
--- a/lib/queryplan-physical-expr.scm
+++ b/lib/queryplan-physical-expr.scm
@@ -603,9 +603,17 @@ partner. */
 		/* A bounded parent probe evaluates this subtree only for rows that survived
 		root braking. Compare those expected probe calls with the dependent stage's
 		input size; retain the group cache when repeated probes amortize its build. */
-		(define inline_presence_stages (if (number? (planner_literal_value decision_probe_work_rows))
+		(define cost_selected_presence_stages (if (number? (planner_literal_value decision_probe_work_rows))
 			(bounded_scalar_query_probe_inline_presence_stages closure_index probe_catalog direct_stages decision_probe_work_rows)
 			'()))
+		/* Cost selection may decline a direct probe, but it cannot turn a stage
+		with residual outer references into a closed initializer. Preserve that
+		logical dependency by evaluating presence in its enclosing row scope. */
+		(define residual_presence_stages (filter nested_stages (lambda (nested_stage)
+			(and (presence_probe_stage? nested_stage)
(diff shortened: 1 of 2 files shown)


carli2:
Added an anonymized production-structure ACL regression in bb3e318dd. It covers repeated request-session EXISTS gates, nested domain-to-parent lookups, CASE/COALESCE/OR, a bounded ID IN list, and query-plan reuse across unrestricted and restricted session bindings. The focused suite now passes 13/13 on the branch. On master it passes 10/13: the minimal residual-OR cause case and both unrestricted ID-projection cases remain red, while the larger structural cases provide broader future coverage.

PR #723, opened 2026-09-01
Stabilize fused mapreducer JIT emission

carli2:
## What
- keep paired Scmer ABI values intact across generated Go calls
- stabilize spill-backed operands, overlapping stack copies, phi frames, and variadic results
- compose forwarding merge callbacks in generated emitters instead of forcing a runtime wrapper
- regenerate the arithmetic emitters needed by fused mapreducers

## Scope
This is a JIT correctness/stability change. It makes arithmetic fused mapreducers compile with no dynamic calls; it does not claim a query-performance result yet.

A real `SELECT SUM(value * 7 + id) ...` run with the patched Go JIT confirms that the filter compiles natively. The complete grouped callback still falls back in `set_assoc_mut` with register pressure, and the shard combiner still falls back in `merge_assoc_mut` on an unsupported stack store. Those complete aggregate subtrees are follow-up work; this PR does not claim they are native yet.

## Verification
- `GOEXPERIMENT=jit GOTOOLCHAIN=local /home/carli/projekte/go-jit/goroot/bin/go test ./scm -run ^TestJITExpressionFusedMapReducerArithmetic$ -count=1 -v`
- `go test ./tools/jitgen ./scm -run ^(TestJITExpressionFusedMapReducerArithmetic|TestNonexistent)$ -count=1`
- patched-JIT server startup: 1276/1276 Scheme unit tests
- SQL result check: filtered arithmetic SUM returned the expected 847; `LogJIT` was used to identify the remaining aggregate fallbacks

### DIFF
--- a/scm/alu.go
+++ b/scm/alu.go
@@ -898,6 +898,11 @@ func init_alu() {
 					ctx.BindReg(result.Reg, &result)
 					ctx.BindReg(result.Reg2, &result)
 				}
+				resultRegsProtected := result.Loc == LocRegPair
+				if resultRegsProtected {
+					ctx.ProtectReg(result.Reg)
+					ctx.ProtectReg(result.Reg2)
+				}
 				lbl0 := ctx.ReserveLabel()
 				bbpos_0_0 := int32(-1)
 				_ = bbpos_0_0
@@ -1329,10 +1334,7 @@ func init_alu() {
 					d30 = JITValueDesc{Loc: LocImm, Type: tagInt, Imm: NewInt(int64(len(args)))}
 					ctx.EnsureDesc(&d2)
 					ctx.EnsureDesc(&d30)
-					ctx.EnsureDesc(&d2)
-					ctx.EnsureDesc(&d30)
-					ctx.EnsureDesc(&d2)
-					ctx.EnsureDesc(&d30)
+					ctx.EnsureDescsTogether(&d2, &d30)
 					var d31 JITValueDesc
 					if d2.Loc == LocImm && d30.Loc == LocImm {
 						d31 = JITValueDesc{Loc: LocImm, Type: tagBool, Imm: NewBool(d2.Imm.Int() == d30.Imm.Int())}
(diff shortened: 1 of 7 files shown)


carli2:
CI analysis: `performance-ab` is red because `tests/performance/wordpress-postmeta-order-limit.yaml` still invokes the handwritten `scan_join_order` shape with the pre-change combine/limit contract. Both sides were reproduced locally with Vanilla Go and their own binary plus `lib/`:

- current `master`: `scan_join_order: combine requires an unlimited scan without offset`
- PR candidate: the identical error

Thus this is not a candidate regression or a measured slowdown. The benchmark itself was not migrated with the scan interface change. We should not restore compatibility with the old scan signature; this performance check needs a one-time override for the transition PR. The normal test workflow remains authoritative and is still running.

carli2:
Fixed in `63f97e5c2`: the handwritten benchmark now uses the fused callback `(lambda (acc ...columns) ...)` and the new optional tail `(neutral combine isOuter notFound batchedProbe)`. Local Vanilla run of `tests/performance/wordpress-postmeta-order-limit.yaml` with `PERF_TEST=1` passes 6/6, including the previously failing case. No performance-check override should be needed after this fix.

carli2:
Manual WordPress benchmark (Ryzen 9 7900X3D, warm caches, exact 8,214/75,000 relevant rows copied from `wordpress_ab_realistic_20260829`, persistent clients, complete result consumption, randomized backend order):

| workload | MariaDB 10.11 median / p95 | MemCP nojit median / p95 | MemCP JIT |
|---|---:|---:|---:|
| point select | 0.0745 / 0.1106 ms (n=600) | 0.3592 / 0.4655 ms (n=600) | 0.4155 / 0.5156 ms (n=1000, paired HTTP run) |
| ORDER BY title LIMIT 5 | 2.3003 / 2.6404 ms (n=400) | 0.5196 / 0.6664 ms (n=400) | correctness failure: `post_status` became `post_stat` |
| filtered postmeta join + ORDER/LIMIT 20 | 0.2774 / 0.3266 ms (n=250) | 0.5341 / 0.6561 ms (n=250) | correctness failure: `NthLocalVar(2) out of range` |

Frontend floors were 0.0463 ms for MariaDB over local persistent MySQL and 0.3587 ms for MemCP over persistent HTTP. A same-protocol Vanilla-vs-JIT point-select run measured 0.3308 vs 0.4155 ms median: JIT is 25.6% slower there. The raw MariaDB-vs-MemCP numbers include the visible protocol/frontend difference.

Additional blocker: the JIT build cannot serve the MySQL frontend; even the first query fails with `Unknown function: session`. Thus the current PR must not claim a complete WordPress JIT speedup. The JIT correctness failures need fixing before ORDER/JOIN speed can be measured.

PR #727, opened 2026-09-02
Remove JIT call frame bridges

carli2:
## What
- remove the remaining `callJIT` and `jitCallSelfCode` Go-frame bridges
- invoke compiled entry points through their actual Go value of type `func(...Scmer) Scmer`
- emit non-tail recursive JIT calls directly through the Go-compatible function value
- keep arguments, entry ownership, and imported-program ownership alive without a deferred trampoline
- retain a benchmark-only model of the removed bridge for direct A/B measurement
- add a recursive panic regression with four consecutive JIT frames

## Runtime prerequisite
Go fork commit `49b59d7737` is already pushed directly to `launix/jit-foreign-frames-go1.27.0`.

The runtime now:
- unwinds arbitrary consecutive registered user frames
- scans and relocates every JIT frame using its precise stack map
- exposes `UnwindDeclare` frames through both `runtime.Stack` and `runtime.CallersFrames` in logical call order
- resolves immutable metadata from the existing return PC only during unwind, GC, profiling, stack growth, or explicit stack inspection

Normal JIT execution performs no descriptor lookup, callback, or extra metadata store.

## Manual A/B
Rebased on current master, same build and benchmark process; 5 samples on AMD Ryzen 9 7900X3D:

- general compiled-procedure adapter: median 17.22 ns/call, 0 B/op, 0 allocs/op
- removed Go-frame bridge: median 10.77 ns/call, 0 B/op, 0 allocs/op
- direct JIT function value: median 9.823 ns/call, 0 B/op, 0 allocs/op
- direct vs bridge: -8.8% callback latency
- direct vs general adapter: -43.0% callback latency

Command: `GOEXPERIMENT=jit ../go-jit/goroot/bin/go test ./scm -run ^$ -bench ^BenchmarkSerialProcJITMapReducerDispatch$ -benchmem -count=5`

## Validation
- rebuilt the canonical patched compiler at `../go-jit/goroot/bin/go`
- complete Go `runtime` suite passed
- nested panic, GC, stack-growth, `runtime.Stack`, and `runtime.CallersFrames` fork tests passed, including 100 repeated runs
- targeted MemCP JIT and vanilla SCM/storage tests passed after rebasing over PR 726
- recursive JIT panic crosses four directly consecutive JIT frames and is recovered by the ordinary Go caller
- WordPress-shaped column-name and direct mapreducer regressions remain green

The pre-existing recursive transferred-merge test failure is reproducible unchanged at merged commit `68c5b894f` with the old Go fork and is unrelated to this change.

### DIFF
--- a/scm/jit.go
+++ b/scm/jit.go
@@ -225,12 +225,10 @@ func (jep *JITEntryPoint) Call(args ...Scmer) (result Scmer) {
 			panic("JIT: invalid hidden argument kind")
 		}
 	}
-	defer func() {
-		runtime.KeepAlive(args)
-		runtime.KeepAlive(jep)
-		runtime.KeepAlive(jep.Owner)
-	}()
-	result = callJIT(jep.Native, args...)
+	result = jep.Native(args...)
+	runtime.KeepAlive(args)
+	runtime.KeepAlive(jep)
+	runtime.KeepAlive(jep.Owner)
 	return result
 }
 
--- a/scm/jit_compile.go
+++ b/scm/jit_compile.go
@@ -469,14 +469,7 @@ func jitApplyCallableSlice(callable, envValue Scmer, args []Scmer) Scmer {
 	return ApplyEx(callable, args, envValue.Any().(*Env))
 }
 
-// jitCallSelfCode preserves a Go unwind frame around a non-tail recursive JIT
-// call. Tail recursion stays inside the generated entry point as a direct jump.
-func jitCallSelfCode(code uintptr, args []Scmer) Scmer {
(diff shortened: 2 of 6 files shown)


carli2:
Status: not ready for review yet. The regular CI test is green, but the direct Go-to-JIT entry locally causes pathological parser execution in the WordPress parser; the same revision returns immediately through the previous Go bridge frame. Heap-copying the arguments, loading the function value into a local, and artificially enlarging the Go caller stack do not change the behavior. I am treating this as a runtime/ABI correctness gap and will not push a bridge or copy workaround.

PR #729, opened 2026-09-02
Stabilize planner and parser JIT control flow

carli2:
## Summary

- spill jitgen's runtime selection from a fixed variadic argument list directly into a stable `Scmer` stack slot instead of protecting every possible source register
- return materialized virtual slices through a frame spill when a Go helper result cannot reserve two registers
- reduce match-list pressure by stabilizing the head before producing the tail and by using the architecture scratch register for slice-capacity arithmetic
- preserve and restore the outer register bank around constant-regex capture emission only when the byte walker cannot reserve its six-register unknown-string state
- regenerate all declaration emitters, including the storage declarations; no `lib/` or test files are changed

This removes the common register-pressure fallback for the requested planner procedures:

| imported procedure | expressions | dynamic calls | native calls | inlined calls |
| --- | ---: | ---: | ---: | ---: |
| `join_order_adaptive` | 309 | 28 | 48 | 17 |
| `join_order_hyperedge_support` | 30 | 0 | 8 | 5 |
| `scalar_query_probe_recipe_plan_using_index` | 172 | 19 | 25 | 13 |
| `lower_direct_boolean_stage_recset_expr` | 111 | 10 | 14 | 12 |
| `build_base_group_scan_assoc_plan` | 138 | 10 | 19 | 9 |
| `scan_join_order_spec` | 467 | 35 | 81 | 20 |
| `sql_queryplan_compile_session` | 110 | 20 | 12 | 1 |

The import dump improves from 35 unique compile failures on `master` to 22. On `master`, six of the seven procedures above still fail; only `lower_direct_boolean_stage_recset_expr` succeeds on its later import pass. All seven select compiled entry points with this change.

## Disassembly review

`sql_queryplan_compile_session` emits 8,651 bytes / 2,211 instructions. Its constant regex is native byte-walker code: the disassembly contains direct byte loads, `byte - '0' <= 9` range checks, and backward repetition branches at offsets `0x8a3` and `0x90f`. It contains no direct target for `regexp`, `Eval`, `Apply`, or `ApplyEx`. The complete function has 40 call instructions and three backward jumps; the dump coverage attributes 20 calls to dynamic Scheme dispatch and 12 to native builtins.

The larger tree walkers are also emitted as single native bodies: `join_order_adaptive` is 35,307 bytes and `scan_join_order_spec` is 68,068 bytes on the final import pass.

## Manual A/B measurements

Baseline: `d493649b6` (`master` after #727 and #728). Candidate: `d77f8391c`. Both were built with `/home/carli/projekte/go-jit/goroot`, `GOEXPERIMENT=jit`, and measured interleaved on the same Ryzen 9 7900X3D host.

| measurement | baseline median | candidate median | change |
| --- | ---: | ---: | ---: |
| warm SQL HTTP execution, `SELECT 1 WHERE "abc" LIKE "a%"`, 12 x 500 | 115.15 us | 110.80 us | -3.8% |
| cold `EXPLAIN COMPILE SELECT 1`, 40 paired samples | 1,001.76 us | 950.40 us | -5.1% |
| cold simple planner time, 80 paired samples | 518.51 us | 517.46 us | -0.2% |
| cold simple HTTP end-to-end, 80 paired samples | 1,231.68 us | 1,227.59 us | -0.3% |
| known map callback, inline microbenchmark, 5 runs | 1,972 ns/op | 1,956 ns/op | -0.8% |
| known map callback, dynamic microbenchmark, 5 runs | 13,020 ns/op | 13,247 ns/op | +1.7% |

Allocations in the map microbenchmark are unchanged: 3 allocs/op inline and 163 allocs/op dynamic.

A 3-10 table join stress query was not used as an A/B result because the `master` baseline itself exceeded the 60-second request deadline. The small SQL measurements above are reported instead of presenting a censored comparison.

## Verification

- `make jitgen` completes in 8.31 seconds and a second run produces an identical generated diff
- `go test ./tools/jitgen -count=1`
- `GOEXPERIMENT=jit go test ./scm -run 'TestJITExpression(Basic|Arithmetic|ListForms|FusedMapReducerArithmetic|RecursivePanicAcrossDirectFrames)|TestJITMatch|TestJITRegex|TestJITParserGrammarMatchesPackrat' -count=1`
- complete `lib/main.scm` startup with JIT dump: 1,276/1,276 Scheme tests and HTTP listener reached

The broader patched-compiler Go run exposes two failures already present on the exact `master` baseline: `TestJITExpressionReduceLambdaArgumentOrder` has the same coverage mismatch, and isolated `TestStorageIntJITEmit` terminates with `unknown caller pc` after #727 because its test helper executes an unregistered raw mmap JIT body directly. This PR does not alter those tests or claim to fix those inherited failures.

## Correctness and executable-memory lifecycle

The follow-up fixes the deterministic correlated-subselect failure by preserving depth-qualified anonymous-lambda captures and reserving hidden capture slots after all numbered locals. It also replaces the recursive pre-entry stack probe with a generated Go-style stack check and precise runtime/jit stack maps.

JIT code now owns an arena lease through its JITEntryPoint. runtime.AddCleanup releases that lease after the Go object becomes unreachable. Full arenas are sealed and are unregistered from runtime/jit and munmaped after their last lease; ShutdownJIT performs the same retirement after server drain, storage shutdown, and the final GC. The current partially filled arena remains reusable instead of being wasted after short-lived query-plan eviction.

Additional verification at head 8d8cba438:

- patched-Go scm package tests: pass
- stock-Go scm package tests: pass
- deleted-row-carrier-visibility.yaml: 14/14
- scan-batch-optimizer.yaml: 12/12, including the formerly failing scalar correlated subselect
- processlist.yaml: 8/8
- SIGTERM shutdown reaches Exit procedure finished after unregistering and unmapping JIT arenas
- GOGC=10 stress testing found and fixed a relocatable scalar-pointer spill omission; further stress investigation remains in progress and is not represented as a completed verification result


carli2:
Intermediate checkpoint `11886bc54` is intentionally pushed before the milestone is green.

- Added seed-growing state for direct and indirect left recursion.
- Emitted rule returns can request a direct jump back to the head rule.
- The former unbounded `290@32 -> 292@32` frame growth now terminates immediately.
- Targeted parser JIT tests and all 1,276 Scheme startup tests pass.
- The three-join reproducer is currently red: rule 312 exposes an incorrect generator/binding value (`ON` instead of the generated join descriptor) during the first growth iteration.

The next checkpoint will fix restoration and publication of generator bindings across seed-growth attempts.

PR #733, opened 2026-09-02
Getting the JIT right

carli2:
## Summary

- make each compiled `Proc` a Go-compatible `func(...Scmer) Scmer` value and prefer its native entry from interpreter and JIT call sites
- store closure captures directly after the compact `Proc` header in a typed `ProcJIT` tail; context-free closures reuse their template, synchronous noescape lambdas use the JIT frame, and escaping lambdas use one typed allocation
- retain the original lambda body and recover it, its native function, and its captures from the concrete `*Proc` in O(1)
- lower static, runtime-known, and self-recursive Proc calls to the compact JIT ABI instead of `ApplyEx`
- preserve lexical scope depth across `begin`, `begin_mut`, `match`, and `match_mut`, including nested scan callbacks
- materialize inline captures when a procedure is closed for persistence, so restart never serializes process-local tail slots
- expose inline captures to scan analysis, preserving point/range index extraction for cached JIT query plans
- keep argument slices borrowed at Proc boundaries; `list` remains the explicit ownership boundary
- use callee-owned split-stack growth, precise JIT stack maps, selective frame initialization, and executable-arena lifetime tracking
- enable the patched-Go JIT correctness workflow for pull requests

## Manual A/B measurements

Vanilla and JIT binaries were built from the same PR source. Vanilla used stock Go 1.24.0. JIT used the `launix-de/go` `jit-foreign-frames-go1.27.0` branch with `GOEXPERIMENT=jit`. Requests used the normal HTTP SQL path, including parser, planner, query-plan cache and execution.

Pinned cold/warm query compilation run:

- server pinned to CPU 2; client pinned to CPU 3
- warm: 20 warmups and 100 byte-identical requests
- cold: 10 warmups and 50 requests with unique, semantically neutral SQL comments
- every workload produced exactly one response digest in both builds

| End-to-end median | Vanilla | JIT | Change |
| --- | ---: | ---: | ---: |
| 4-table join, cold | 32.692 ms | 11.700 ms | 2.79x faster |
| simple indexed select, cold | 3.885 ms | 3.131 ms | 1.24x faster |
| 4-table join, warm cached Proc | 328.461 us | 304.897 us | 1.08x faster |
| simple indexed select, warm cached Proc | 312.531 us | 311.148 us | neutral |

After fixing captured constants in scan analysis, a separate 10,000-row warm run used 30 warmups and 500 samples over a persistent HTTP connection (5% trimmed mean):

| Warm end-to-end SQL | Vanilla | JIT | Change |
| --- | ---: | ---: | ---: |
| primary-key point select | 380.37 us | 352.88 us | 1.08x faster |
| indexed range plus `SUM` | 579.95 us | 368.42 us | 1.57x faster |
| filter, projection and ordered limit | 678.49 us | 641.26 us | 1.06x faster |

The sub-millisecond point and projection cases are dominated by HTTP, transaction and SQL dispatch. The cold gain measures the imported SQL compiler/planner executing as native code; the warm range gain includes the cached query plan and its native scan/mapreducer callbacks.

## Disassembly and code-generation audit

`MEMCP_JIT_DUMP_DIR` was sampled at startup, after one cold four-table request, and after the identical warm request.

- the cold request produced the cached plan and closure templates once; the warm cache hit emitted no additional code
- representative generated query-tree bodies were 19,663 bytes with 90 native/helper calls and 4,077 bytes with 21 calls
- ordinary query lambdas skip blanket frame clearing; parser control-flow frames retain it for branch-dependent Phi homes
- direct recursive calls reload the current Go funcval and call its code pointer, preserving the concrete typed closure context
- runtime symbol resolution uses the `Env` held by the concrete Proc; `with_session` changes that context without recompilation
- the imported `join_order_adaptive` remains a large native body rather than an interpreted outer wrapper; its audited 147 calls comprised 29 compact/direct calls and 118 Go-ABI builtin/helper boundaries, not `Eval` fallbacks

## Correctness

- `GOEXPERIMENT=jit go test ./scm`: passed (393 Go tests)
- `GOEXPERIMENT=jit go test ./tools/jitgen`: passed
- `tests/execution/operators/range-scan-coverage.yaml`: 113/113 passed, including point probes, nested pushdown and post-`SHUTDOWN` queries
- targeted restart regression: 5/5 passed; persisted callbacks no longer retain process-local JIT captures
- `tests/sql/expressions/prepared-stmts.yaml`: 57/57 passed
- `tests/storage/persistence/incremental-invalidation.yaml`: 33/33 passed
- `tests/storage/indexes/fulltext-like-index.yaml`: 86/86 passed, including LIKE+Range
- context-free, captured, nested-scope, recursive, `with_session`, parser, analyzer and interpreter-to-JIT Proc boundaries have dedicated regression coverage
- JIT pull-request CI is enabled by this PR

The patched Go branch contains `runtime/jit.TailTypeFor[T]` and its repeated-tail GC hardening, so MemCP does not link directly to `runtime.mallocgc` or use reflection in the closure hot path.


carli2:
I isolated one independent JIT correctness failure and opened #735 as a small follow-up against this branch. Large parser procedures could lose a selected `coalesce` result before `setN` consumed it. The fix changes the coalesce phi to a rooted frame slot. It moves `tests/rdf/rdf-order-limit.yaml` from 4/9 to 9/9 under JIT.

The current LIKE+Range speedup gate itself is green, so that earlier suspected regression is no longer outstanding.

carli2:
Two focused follow-up fixes are ready:

- #735 fixes JIT coalesce branch-result corruption under register pressure. It restores `rdf-order-limit.yaml` from 4/9 to 9/9 under JIT.
- #736 passes the recursive RDF `build_scan` closure explicitly into the property-path helper. It restores `rdf-advanced-sparql.yaml` from 7/9 to 9/9 under JIT, while both targeted vanilla suites and the full `make test` run pass.

The reported LIKE+Range regression is not present in the latest #733 JIT log: `LIKE+Range speedup >= 3x` passes there.

carli2:
A third independent follow-up is ready in #737. JITGen was flattening a helper that consumes a two-result Go call into a nested CFG and lost the `(matched, handled)` result. This made dynamic JIT `strlike` return false and caused the PostgreSQL `ILIKE` + `::text` correctness failure.

#737 keeps tuple-consuming helpers at one static native Go call boundary (not Scheme dispatch), shrinks the generated `strlike` emitter by roughly 700 lines, and moves `tests/sql/expressions/functions.yaml` from 50/51 to 51/51 under JIT. Patched-Go `go test ./scm`, generator tests, and the full vanilla `make test` pass.

PR #736, opened 2026-09-03
planner: pass recursive RDF scan builder explicitly

carli2:
## Summary

- pass the recursive `build_scan` planner closure explicitly into the RDF property-path helper
- avoid capturing a later lexical binding as `nil` when the parser is JIT-compiled
- keep the change at planning time, without adding work to query scan loops

## Verification

- full `make test` hook-equivalent suite passes
- JIT `tests/rdf/rdf-advanced-sparql.yaml`: 7/9 -> 9/9
- JIT `tests/rdf/rdf-queryplan-coverage.yaml`: 9/9
- vanilla versions of both suites: 9/9

### DIFF
--- a/lib/rdf-parser.scm
+++ b/lib/rdf-parser.scm
@@ -689,7 +689,7 @@ consumer stage. */
 			(list? l) (eval l)
 			(error "SPARQL error: unsupported property path subject " expr)
 		)))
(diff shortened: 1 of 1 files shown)


carli2:
Targeted CI confirmation: both affected JIT suites are now green (`Advanced SPARQL` 9/9 and the separate property-path coverage 9/9). The overall job remains red only because this stacked PR intentionally does not include the other independent #733 JIT fixes.

PR #756, opened 2026-09-05
JIT-compile trigger procedures

carli2:
## Summary

- optimize and evaluate deferred trigger plans through the same phase order as cached SQL query plans
- vectorize eligible maintenance triggers before native compilation
- recursively JIT-compile trigger orchestration and its hot scan/filter/map/reduce callbacks
- preserve public lambda slots when raw internal ASTs omit optimizer metadata
- return the original procedure automatically for unsupported expressions and non-JIT builds
- make Scheme-language trigger sources deferred so they receive the same optimize/evaluate/JIT pipeline after restore

## Why

Trigger bodies were retained as reusable `Proc` values, but firing a trigger called those procedures through the interpreter. `scm.Apply` already dispatches a Proc with `JITCode` directly to its native entry point, so no trigger execution API or persistence format needs to change.

The phase order is important: vectorization first needs the retained Scheme body, while JIT requires optimizer-normalized numbered locals. Trigger compilation now also supports raw internal maintenance ASTs. Such nested lambdas do not carry an explicit `numVars` field; their public parameters nevertheless occupy the leading numbered slots. Previously recursive compilation started captures at slot zero, so the first scan argument overwrote the captured `OLD` or `NEW` row. That silently prevented keytable and aggregate-cache maintenance filters from matching. The general lambda compiler now derives the minimum local count from the declared parameters before assigning capture slots.

The earlier workaround that excluded hidden physical triggers has been removed. Both generated and Go-created maintenance triggers remain eligible for recursive JIT compilation, while an already vectorized trigger keeps its native batch implementation.

## Manual A/B measurement

Same patched-Go JIT build, same Proc and arguments, five samples per variant. The measured trigger body reads `NEW.value` and adds one:

- interpreter baseline: 449.6–522.3 ns/op, 576 B/op, 5 allocs/op
- JIT trigger: 35.85–36.64 ns/op, 0 B/op, 0 allocs/op
- improvement: approximately 12.3x–14.6x faster with all per-call allocations removed from the measured Proc dispatch/body

This isolates the trigger Proc improvement. Row-dictionary construction, transaction savepoints, and the DML performed by a trigger are outside this microbenchmark and remain unchanged.

## Cached-plan ownership fix

`CREATE TRIGGER` statements are query-plan cacheable. The optimizer rewrites local-variable references in place, so trigger compilation now clones the retained AST before optimization. Without that ownership transfer, dropping and recreating the same scalar-subquery trigger could optimize already-numbered locals and fail with an out-of-range local slot.

## Validation

- compiler regression test reproduces the raw scan-shaped callback capture collision before the fix and passes 20/20 after it
- patched-Go JIT unit tests prove native entry points for direct, lazy/restored, internal, and hidden physical triggers
- JIT keytable COUNT: 15/15, 99 ms
- JIT FK-to-PK group-cache reuse including INSERT, DELETE, and UPDATE: 21/21, 85 ms
- JIT incremental SUM/COUNT deletion, partitioned ORC invalidation, and concurrent maintenance: 33/33, 196 ms
- exact AFTER INSERT scalar self-lookup and cached trigger recreation: 10/10, 63 ms
- deferred trigger compile-error recovery: 8/8, 44 ms
- `lib/main.scm` passes the Scheme formatter check

The complete test matrix is left to CI as requested.

### DIFF
--- a/lib/main.scm
+++ b/lib/main.scm
@@ -43,7 +43,8 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 available without loading an SQL module; other frontends can register their
 own compiler under any language name. */
 (registertriggerlanguage "scheme" (lambda (source context)
-	(eval (scheme source (concat "trigger:" (context "schema") "." (context "table") ":" (context "name"))))))
+	(list (quote deferred_trigger)
+		(scheme source (concat "trigger:" (context "schema") "." (context "table") ":" (context "name"))))))
 
 (import "sql.scm")
 (import "dashboard.scm")
--- a/scm/jit.go
+++ b/scm/jit.go
@@ -7448,6 +7448,14 @@ func jitCompile(a ...Scmer) Scmer {
 	return jitCompileMode(true, a...)
 }
 
+// CompileJIT compiles a procedure to native code when this build enables the
+// JIT. recursiveLambdas controls whether nested procedures are compiled with
+// their parent. Unsupported procedures and builds without JIT support retain
+// the original procedure as their interpreter fallback.
+func CompileJIT(procedure Scmer, recursiveLambdas bool) Scmer {
+	return jitCompileMode(recursiveLambdas, procedure)
+}
+
 func jitCompileMode(recursiveLambdas bool, a ...Scmer) Scmer {
 	return jitCompileModePublish(recursiveLambdas, true, true, a...)
 }
--- a/scm/jit_expression_test.go
+++ b/scm/jit_expression_test.go
@@ -130,6 +130,56 @@ func TestJITFunctionValueCarriesOriginalProcAndInlineCaptures(t *testing.T) {
 	}
 }
 
+func TestJITScanShapedMaintenanceCallbacksCaptureOuterRow(t *testing.T) {
+	const name = "jit_test_scan_shaped_maintenance_callbacks"
+	called := false
+	params := []*TypeDescriptor{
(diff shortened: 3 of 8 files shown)


carli2:
CI follow-up (`670ce4882`): the JIT failure was caused by compiling hidden physical maintenance triggers as one additional outer native procedure. Those generated plans already own JIT boundaries at their scan/DML callbacks; compiling the large orchestration duplicated that work during lazy group-cache creation and caused the aggregate suites to time out after 50 seconds, followed by server restarts.

The corrected boundary is structural rather than a query-shape special case:

- visible language-defined triggers: compile outer Proc
- compact Go-created `IsSystem` triggers: compile outer Proc
- hidden relational-maintenance plans: retain interpreter orchestration and callback-owned JIT boundaries

Focused patched-Go JIT validation after the fix:

- `count-where-keytable.yaml`: 15/15, 115 ms (CI failure previously timed out at 50.653 s)
- `fk-pk-group-reuse.yaml`: 21/21, 103 ms (CI failure previously timed out at 50.623 s)
- `trigger-after-insert-self-lookup.yaml`: 10/10, 70 ms
- `trigger-body-compile-error.yaml`: 8/8, 267 ms
- trigger JIT/ownership Go tests: pass

Normal full CI and performance A/B were already green on the preceding rebased head. A fresh JIT CI run is now active for this fix.

PR #771, opened 2026-09-06
Optimize composite scan lookups

carli2:
## Summary

- bind canonical composite lookup values directly instead of decoding every boundary for every candidate row
- keep shard callback state in one heap object and use stack-backed column-reader arrays for up to eight match columns
- preserve fast composite unique-key detection when planner boundary order differs from index declaration order
- add absolute allocation limits, permanent dimension benchmarks, three SQL performance carriers, physical-plan assertions, and an authenticated non-admin point-query performance case

## Manual A/B measurements

Baseline: current master at 28b9cb6f1. Candidate: 158447aad. Both used the same 1,024-row in-memory fixture, one warmup call per benchmark process, and five benchmark samples on an AMD Ryzen 9 7900X3D. The composite unique key deliberately reverses the planner boundary order.

| Warm compiled operator | Master median | Candidate median | Change | Master allocs / bytes | Candidate allocs / bytes |
| --- | ---: | ---: | ---: | ---: | ---: |
| two-column value lookup | 2,874 ns | 2,404 ns | -16.4% | 8 / 296 B | 3 / 240 B |
| two-column exists lookup | 2,816 ns | 2,459 ns | -12.7% | 8 / 296 B | 3 / 240 B |
| two-column map lookup | 3,015 ns | 2,473 ns | -18.0% | 10 / 336 B | 5 / 272 B |

Command: `go test ./storage -run "^$" -bench "^BenchmarkCompiledScanLookupDimensions/(value_two|exists_two|map_two)$" -benchmem -count=5`

The SQL carrier A/B executes 20,000 composite probes per query with three warmups and seven samples. Manual runs measured value 269.9 -> 269.1 ms (-0.3%), map 270.1 -> 271.6 ms (+0.6%), and exists 256.9 -> 259.7 ms (+1.1%). These query-level differences are within local run-to-run noise; the cases primarily provide CI A/B regression coverage around the complete planner and interpreter path.

The authenticated HTTP point-query case used identical setup, 10 warmups, and 100 samples per run across five fresh server runs. Its local median was 633 us on master and 649 us on the candidate (+2.4%), within the observed 610-684 us run-to-run range. The approximately 0.3 us operator saving is therefore not reliably resolvable through local HTTP noise.

The final CI A/B run measured every added case positively: authenticated point lookup 1.207 -> 1.166 ms (-3.4%), composite value 467.483 -> 464.594 ms (-0.6%), composite map 474.717 -> 471.415 ms (-0.7%), and composite exists 452.524 -> 449.586 ms (-0.6%). CI is additional evidence, not a substitute for the manual measurements above.

## Verification

- `go test ./storage`
- `go test -race ./storage -run "^Test(ScanLookup|CompiledScanLookup)" -count=1`
- `PERF_TEST=1 python3 run_sql_tests.py tests/performance/authenticated-point-lookup.yaml --log-times`
- full repository test suite delegated to CI as required

### DIFF
--- a/git-pre-commit
+++ b/git-pre-commit
@@ -305,8 +305,10 @@ run_memcp_forever() {
     if [ $rc -ne 0 ]; then
       echo "⚠️  MemCP exited unexpectedly (code=$rc). Restarting..."
       if [ -f "$memcp_log" ]; then
-        echo "--- MemCP log at unexpected exit (last 200 lines) ---"
-        tail -n 200 "$memcp_log"
+        echo "--- MemCP log at unexpected exit (last 2000 lines) ---"
+        # Runtime fatal reports include every goroutine. Keep enough context to
+        # retain the actual fault header instead of only its trailing stacks.
+        tail -n 2000 "$memcp_log"
         echo "--- end MemCP log at unexpected exit ---"
       fi
     fi
--- a/storage/scan.go
+++ b/storage/scan.go
@@ -1361,6 +1361,24 @@ func (t *table) hasBoundUniquePoint(access scanAccess) bool {
 			if matched {
 				return true
 			}
+			covered := len(unique.Cols) > 0
+			for _, column := range unique.Cols {
+				matched = false
+				for i := 0; i < access.len(); i++ {
+					if access.boundaryColumn(i) == column && !access.values[i].IsNil() {
+						matched = true
+						break
+					}
+				}
+				if !matched {
+					covered = false
+					break
+				}
+			}
+			if covered {
+				return true
+			}
+			continue
 		}
 		covered := true
 		for _, col := range unique.Cols {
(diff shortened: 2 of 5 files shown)


carli2:
JIT CI follow-up:

- Investigated the failed run 34026752478: the reported `nested-exists-not-aggregate.yaml` setup timeout was collateral damage from two earlier MemCP exits with code 2 while unwinding through active JIT frames.
- Merged current master, including #772’s raw parser-state JIT helper path, and retained 2,000 crash-log lines so future runtime failures preserve their fatal header.
- Exact patched-Go local stress checks passed with `GOEXPERIMENT=jit`, `GOGC=1`, and 20 repetitions of the parser/stack/unwind/panic/frame JIT tests.
- Full JIT CI run 34030738744 passed, including dump/restore. Its log contains no code-2 exit, fatal error, unknown caller PC, or unexpected return PC. The sole code-138 restart belongs to the explicit managed-restart test.

PR #774, opened 2026-09-06
Plan weighted JIT register slots offline

carli2:
## Problem

The one-pass JIT cannot cheaply reconstruct loop liveness while emitting machine code. Generated storage-reader loops therefore reload scalar loop phis from stack slots. Building SSA or coloring at query-compile time would defeat the one-pass design, while hard-coding amd64 registers in jitgen would break architecture independence and nested emission.

## Solution

This PR splits allocation into two stages:

1. **Offline in jitgen:** calculate loop-phi liveness, interference, coloring, and reload weights; emit a compact architecture-neutral slot plan.
2. **During JIT emission:** map the most valuable slots onto registers exposed by the backend while retaining its temporary-register reserve. Unmapped values keep their canonical stack homes.

The CI investigation established the ownership boundary mechanically. A generated builtin may use its offline plan only when its SSA graph contains every environment that can be live during the loop. Leaf builtins and standalone storage readers satisfy that rule: runtime allocation preserves caller-live homes and storage readers additionally have an explicit frame boundary. A builtin containing a dynamic Scheme callback does not satisfy it, because the surrounding Proc compiles that callback outside the builtin's standalone SSA graph. `inlineRegisterPlanSafe` detects that structural boundary and leaves those loops on their established stack homes until a joint caller/callback interference graph exists. This is based on call-graph shape, not a list of operator names.

The distinction fixes both failure modes found by CI: independently planning callback loops corrupted JSON aggregate state, while treating all ordinary plans as merely "dormant" changed their stack-home codegen and made complex derived-table queries stop making progress. Runtime symbol lookups also produce into a temporary pair while the planned destination remains pinned, so Go ABI result allocation cannot overwrite the destination or its arguments.

Scmer, strings, and slices are not colored as indivisible wide values. Their words have different pointer/liveness properties and type folding can eliminate words. They retain the established stack representation until split-word liveness is modeled.

Register-plan connected components are sorted before coloring, making repeated `make jitgen` output deterministic.

## Manual measurements

Initial measurements on an AMD Ryzen 9 7900X3D, CPU 2 pinned, patched Go with `GOEXPERIMENT=jit`, identical 128-element fixtures, 600 ms per sample, seven samples (medians):

| Generated loop | master | independently activated plan | change |
|---|---:|---:|---:|
| `for` | 10,736 ns/op | 10,619 ns/op | 1.1% faster |
| `map` | 2,207 ns/op | 2,166 ns/op | 1.9% faster |
| `map_mut` | 1,709 ns/op | 1,664 ns/op | 2.6% faster |
| `reduce` | 1,122 ns/op | 1,098 ns/op | 2.1% faster |
| `filter` | 3,855 ns/op | 3,818 ns/op | 1.0% faster |
| `filter_mut` | 1,868 ns/op | 1,908 ns/op | 2.1% slower |

The callback-bearing measurements quantify the opportunity for the joint-plan follow-up; they are not claimed as a final runtime win here. Closed leaf loops and standalone storage readers can use the allocation infrastructure now.

## Correctness verification

Targeted patched-Go/JIT suites after the ownership fix:

- MySQL JSON functions: 53/53
- PostgreSQL JSON functions: 56/56
- composed correlated scalar aggregates: 25/25
- `IN`/`NOT IN` subqueries including `UNION ORDER BY/LIMIT`: 73/73
- column collation across restart: 9/9
- badge scalar aggregates: 2/2
- name-resolution scopes: 40/40
- window functions: 54/54
- campaign-style derived aggregate regression: 3/3 (118 ms locally; previously timed out)
- application derived-table regressions: 4/4 (163 ms locally; previously timed out)
- focused JIT register/lookup Go tests
- `go test ./tools/jitgen`
- repeated `make jitgen` produces an identical diff

The complete SQL/JIT and performance matrices run in CI.

## Follow-up

- Build a joint caller/callback interference graph before planning dynamic-callback loops.
- Represent Scmer words independently so pointer/tag/payload lifetimes can diverge safely.
- Apply the standalone allocator to specialized storage-reader loops.


carli2:
Fix pushed in `a3ac690ef`.

The failing badge aggregate suite was reproduced locally under `GOEXPERIMENT=jit`. The regression came from routing runtime/global symbol resolution directly into a register-planned result pair: that API did not establish the non-aliasing required across the surrounding inline environment, and compound filter terms were corrupted. Symbol lookup now resolves through temporary ABI result registers and is placed into the requested pair only afterwards.

The same commit also centralizes scalar/pair phi representation conversion in `EmitStoreScmerToStack`; jitgen no longer assembles `Scmer` words manually. This removes about 1.7k generated lines and avoids malformed bool/int/float/nil representations.

Targeted validation:
- normal build: pass
- JIT build with patched GOROOT: pass
- `go test ./tools/jitgen`: pass
- focused JIT register/global-call tests: pass
- `tests/planner/aggregates/badge-scalar-aggregates.yaml`: 2/2 pass (previously 0/2 with exact CI mismatch)

Full CI is running now. Joint caller/callee register allocation remains a follow-up; this PR deliberately retains the documented conservative inline boundary.

carli2:
Follow-up correction pushed in `099bb1020`.

The next JIT CI run exposed a deeper failure hidden behind the original badge mismatch: heap corruption in the Go runtime during the deep name-resolution/grammar test. Bisecting the PR showed the two existing alternatives were both invalid at the same inline boundary:

- direct lookup result placement: corrupts live compound-filter values;
- temporary lookup result without protecting the requested target: allows result allocation to evict/reassign the target, later writing through stale register identities.

The fixed invariant is now explicit in `jitEmitGoCallPairViaTemporary`: pin both requested pair registers, emit the lookup into temporary ABI result registers, perform the final pair placement while the target is stable, then unpin. I also fully reverted my overly broad generated-phi cleanup from the previous follow-up; `isPhiPairType` covers more than `Scmer`, so that generalization was unsafe.

Focused JIT validation on the final code:
- deep name-resolution/grammar: 40/40
- badge scalar aggregates: 2/2
- EXISTS/session-variable shapes: 8/8
- window functions: 54/54

The new full CI run is pending.

PR #793, opened 2026-09-07
Harden transactions against persistence failures

carli2:
## Summary

- add deterministic, environment-scoped persistence fault injection for before-call and real partial stream/WAL writes
- standardize backend I/O failures as `*PersistenceFailure` and log them directly to stderr without depending on `system_statistic.errors`
- roll back definite pre-authority failures in memory without issuing more failing I/O, and keep explicit transactions aborted until `ROLLBACK`
- distinguish failed post-authority durability barriers as commit-outcome-unknown instead of performing an unsafe in-memory rollback
- bound S3 SDK attempts and Ceph retries to three attempts
- replace the WAL `Sync()` contract with `Flush(durable bool)`, so `logged` S3/Ceph transactions transmit buffered WAL data without forcing local-filesystem fsync semantics
- make filesystem WAL frame writes recover their prior offset and length after short/partial writes

The common single-shard commit path still performs one flush/durability barrier, not two. Multi-shard commits retain the required prepare barriers before publishing coordinator authority.

## Reliability drill

The opt-in `tools/reliability_drill.py --mode io-failures` suite was added test-first. On the baseline, the partial-WAL scenario failed because the transaction incorrectly returned success. During implementation it also exposed a one-row in-memory rollback leak and an explicit transaction that accepted later statements after being aborted; both are now covered.

Final deterministic run (`--io-failure-rows 201 --seed 42`) passed three scenarios:

1. partial WAL failure in an explicit ACID transaction, aborted-state enforcement, `ROLLBACK`, and successful retry
2. partial WAL failure in autocommit, exact rollback, successful retry, and crash recovery
3. failed pre-authority sync for a multi-shard ACID update, exact rollback, successful retry, and crash recovery

## Local validation

- `go test ./storage ./scm`
- `go test -tags ceph ./storage`
- `python3 -m unittest tools/test_reliability_drill.py`
- `go build -o memcp .`
- `python3 tools/reliability_drill.py --mode io-failures --io-failure-rows 201 --seed 42 --artifacts .reliability-final`

The full repository suite is intentionally delegated to CI per the repository policy.

## Remaining policy decision

A failure after commit authority has been accepted cannot honestly be reported as a definite rollback: an fsync timeout or remote response loss may still leave the marker durable. This PR reports that case as `commit outcome unknown` and preserves committed in-memory visibility to avoid immediate memory/recovery divergence. A later design can add an isolated SCM critical-event hook; it should be asynchronous, bounded, time-limited, independent of database persistence, and unable to alter commit semantics.

## Performance A/B

CI compared the final candidate against the exact PR base with identical fixtures and runner settings. The previously variable cold aggregate-pushdown case measured 85.070 ms -> 85.375 ms (+0.4%); its warm counterpart measured 5.278 ms -> 5.196 ms (-1.6%). The complete Performance A/B job passed.

### DIFF
--- a/storage/database.go
+++ b/storage/database.go
@@ -129,21 +129,28 @@ func (db *database) transactionCommitted(txID string) bool {
 	return committed
 }
 
-func (db *database) commitTransaction(txID string, durable bool) {
+func (db *database) commitTransaction(txID string, durable bool) error {
 	if txID == "" {
-		return
+		return nil
 	}
 	db.initializeTransactionLog()
 	db.transactionMu.Lock()
 	defer db.transactionMu.Unlock()
 	if _, exists := db.committedTx[txID]; exists {
-		return
+		return nil
 	}
-	db.transactionLog.Write(LogEntryCommit{txID: txID})
-	if durable {
-		db.transactionLog.Sync()
+	if err := persistenceCall(func() { db.transactionLog.Write(LogEntryCommit{txID: txID}) }); err != nil {
+		return err
+	}
+	if err := persistenceCall(func() { db.transactionLog.Flush(durable) }); err != nil {
+		// The authority frame was accepted into the logfile. A failed local
+		// durability barrier or remote publication cannot tell us whether it
+		// will survive a crash, so memory must keep treating it as committed.
+		db.committedTx[txID] = db.transactionGen
+		return markCommitOutcomeUnknown(err)
 	}
 	db.committedTx[txID] = db.transactionGen
+	return nil
 }
 
 func (db *database) transactionCompactionSnapshot() (uint64, bool) {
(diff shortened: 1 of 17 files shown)


carli2:
The first full-test run exposed a real rebuild-generation regression: visibility propagation crossed a retired intermediate generation and attempted to write its already closed WAL. Previously that backend error was silently ignored. The follow-up commit keeps in-memory RecID/visibility propagation across the complete chain but persists only into the current generation.

Targeted validation after the fix: `tests/execution/concurrency/transactions.yaml` passes 204/204, including cursor rollback and ACID commit across two rebuilds. Storage, SCM, and Ceph-tagged Go tests also pass locally.

carli2:
Storage interface audit follow-up: WalkBlobs and WalkShardFiles no longer expose an ignored backend-error return. Their callbacks are now non-failing, and filesystem, S3, and Ceph traversal failures consistently panic with *PersistenceFailure, matching every other persistence boundary. Missing filesystem blob storage remains the explicit empty-store case; ENOTDIR and other I/O failures are covered by a new unit test.

Revalidated locally on commit 0c47b89f9:
- go test ./storage ./scm
- go test -tags ceph ./storage
- python3 -m unittest tools/test_reliability_drill.py
- I/O failure drill: 3/3 scenarios passed with a 201-row multi-shard fixture and seed 42

PR #799, opened 2026-09-07
Move database storage with ALTER DATABASE

carli2:
## Summary

- add `ALTER DATABASE ... SET STORAGE` for explicit backend options
- add `FROM DATABASE` and `FROM TABLE` forms that reuse connection settings while deriving an isolated target prefix
- build private shard successors through the existing rebuild-generation chain while readers stay on the published topology
- mirror concurrent writes into those successors, convert their WALs to the destination-native format, and publish all table topologies only after the destination schema is durable
- serialize only rare catalog membership changes; scan, lookup, DML, and the idempotent planner `createtable` fast path do not enter the migration mutex
- use the shared immutable `ReadMap[K,V]` from [NonLockingReadMap v1.1.1](https://github.com/launix-de/NonLockingReadMap/releases/tag/v1.1.1) for both database and table catalogs
- store catalog keys separately from mutable database/table values, eliminating value copies and local duplicate map implementations
- prefer a backend descriptor over a legacy same-name filesystem directory during restart

## Safety semantics

- destination schema is the commit record and is published only after columns, blobs, shard WALs, and the transaction commit WAL
- any failure before publication clears rebuild forwarding and leaves the old topology and source backend authoritative
- concurrent writes made during the copy are retained in the prepared generation and survive a destination-only reload
- old readers and transactions keep the retired source generation pinned; source cleanup starts only after its existing generation drain signal closes
- DDL/catalog writers wait for migration, while ordinary readers continue on the old generation
- catalog readers observe atomically published immutable membership snapshots and never acquire the writer mutex
- a destination that already publishes a schema is rejected

## Tests

- `go test ./...`
- targeted `-race` tests for online reader continuity, concurrent write replay, catalog-writer exclusion, copier failure, and cutover/reload
- upstream NonLockingReadMap normal and full race test suites
- immutable old-snapshot, JSON round-trip, ordering, concurrent reader/writer, and zero-allocation lookup coverage upstream
- `tests/sql/compatibility/dbcheck-ddl-compat.yaml` coverage for explicit and copied storage configurations
- local MinIO storage migration coverage from the preceding storage compatibility work
- CI runs the full SQL, JIT, packaging, alternative-storage, and performance A/B jobs

## Manual performance A/B

Same machine, fixture, warmup, and benchmark configuration.

Database catalog lookup:

- before (`9f2d2b78d`, object-keyed map): 21.31-23.90 ns/op, 8 B/op, 1 alloc/op
- shared `ReadMap`: 2.37-2.59 ns/op, 0 B/op, 0 allocs/op
- improvement: about 89% lower lookup latency and one allocation removed

Table catalog lookup:

- before (`168042771`, object-keyed map): 32.11-34.75 ns/op, 8 B/op, 1 alloc/op
- shared `ReadMap`: 2.59-2.60 ns/op, 0 B/op, 0 allocs/op
- improvement: about 92% lower lookup latency and one allocation removed

The specialized local database catalog briefly measured 1.96-2.02 ns/op. The shared generic implementation costs roughly 0.4-0.6 ns for a one-entry catalog while removing the duplicate implementation and accelerating every table lookup. The scan/DML hot path contains no `storageMoveMu` acquisition. During a migration, a deliberately blocked destination copy was verified to leave the exact old topology pointer published while point lookup completed within the test deadline.


carli2:
converted to draft because it intruduces global locks on readers for each table. can be converted again when the issue is fixed.

PR #809, opened 2026-09-07
JIT: propagate finalized main-storage value types

carli2:
## Problem

A storage-specialized JIT reader currently returns a dynamically typed `JITValueDesc` even when the finalized main storage proves that every row has one physical `Scmer` tag. The immediately composed filter/map/reduce emitter therefore retains generic tag dispatch and, for some declarations, a generic argument-slice call boundary.

This information is valid for immutable main storage only. Delta rows can have a different runtime tag and must remain dynamically typed.

## Solution

- add `ColumnStorage.JITValueType()` as a conservative exact-tag contract for finalized main storage
- implement the contract for all storage implementations, returning `JITTypeUnknown` for nullable or mutable/variant representations
- retain the float NULL proof during build and reconstruct it during deserialization, so querying the type remains O(1)
- feed the proven type into the main-only fused storage consumer
- leave delta lookup and callback dispatch unchanged
- add storage-shape and delta-regression tests

`StorageSCMER` and compute proxies deliberately stay unknown because their values can be repaired or replaced after the analysis pass. String compression reports its actual physical `String`, `CString`, or `BString` tag only when empty/NULL representation cannot vary it.

## Manual A/B measurement

Same checkout, same generated storage reader, fixed CPU 2, `GOMAXPROCS=1`, patched Go with `GOEXPERIMENT=jit`, 60,000 integer values followed immediately by the Scheme predicate `(< value 511)`, 500 ms samples, 10 repetitions. The only A/B variable is the descriptor that the consumer receives: the previous unknown tag versus the finalized storage's exact tag.

| Variant | Median | Allocations | Change |
|---|---:|---:|---:|
| unknown type (baseline) | 737,247 ns / 60k rows | 0 | baseline |
| finalized main type | 605,174 ns / 60k rows | 0 | **17.9% faster** |

Command:

```sh
GOMAXPROCS=1 taskset -c 2 env GOROOT=../go-jit/goroot GOEXPERIMENT=jit ../go-jit/goroot/bin/go test ./storage -run '^$' -bench '^BenchmarkStorageIntTypedConsumer$' -benchtime=500ms -count=10
```

## Disassembly

For the same composed reader/predicate:

- unknown type: 399 bytes, `0x50` local frame, generic two-`Scmer` declaration boundary
- exact main type: 372 bytes, `0x20` local frame, direct typed declaration path

This is **27 bytes / 6.8% smaller** and removes the generic argument materialization boundary. The storage bit-unpacking body is otherwise identical, so the measurement isolates type propagation rather than a different reader.

## Validation

- `go test ./...`
- `GOROOT=../go-jit/goroot GOEXPERIMENT=jit go test ./scm ./storage ./tools/jitgen`
- post-rebase `go test ./storage` with both stock Go and patched JIT Go
- JIT SQL needle: `tests/sql/expressions/basic-sql.yaml`, 33/33 passed

## Follow-up

Use this contract while emitting the query-specific main-storage scan pipeline so known filter/map/reduce bodies can fold type branches together with the storage constants. Delta processing remains a separate dynamically typed tail of the scan.

### DIFF
--- a/storage/shard_jit.go
+++ b/storage/shard_jit.go
@@ -225,7 +225,7 @@ func emitFusedLoop(ctx *scm.JITContext, mainCols []ColumnStorage, mapProc, reduc
 		idxCopy := ctx.AllocReg()
 		ctx.EmitMovRegReg(idxCopy, recidReg)
 		idxDesc := scm.JITValueDesc{Loc: scm.LocReg, Type: scm.TagInt, Reg: idxCopy}
-		colDescs[i] = col.JITEmit(ctx, idxDesc, scm.JITValueDesc{Loc: scm.LocAny})
+		colDescs[i] = emitMainStorageValue(ctx, col, idxDesc, scm.JITValueDesc{Loc: scm.LocAny})
 	}
 	ctx.UnprotectReg(recidReg)
 	ctx.FreeReg(recidReg)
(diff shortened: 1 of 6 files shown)


carli2:
CI failure analysis update:

- The failed JIT run returned both `A` and `B` for the cold and warm SPARQL MINUS checks.
- A retry of the exact same commit reached the same suite before it was superseded and passed all three MINUS checks, including cold and warm execution.
- This is not an exact query-plan-cache key collision: SPARQL literal parameterization is disabled, so the cache key contains the FNV-1a hash of the complete query. I checked all 8,074 SQL/SPARQL query strings in `tests/`; there are no distinct-query hash collisions, and this query's hash is unique.
- The production behavior added by this PR is not active in that test: exact main-storage types enter consumers only through `compileFusedMainLoop`, while `enableFusedMainJIT` is currently `false`. The live `StorageFloat.hasNull` field is bookkeeping whose only reader is that inactive path/tests.

The branch is now rebased onto current `master` (`eb0a612f8`), patched-Go `go test ./scm ./storage` passes, and the retained mixed/rebuilt RDF data directory passes `rdf-spec11-algebra-complete.yaml` 21/21. The fresh post-rebase CI run is the acceptance run; no feature was disabled and no workaround was added.

PR #810, opened 2026-09-07
Add streaming HTTP response proxy

carli2:
Adds a response-local `proxy` primitive for trusted Scheme HTTP handlers. It forwards the inbound method, body, and headers to a target URL and streams the upstream status, headers, and body back to the client. This lets server-rendered applications keep loopback-only services such as Ollama behind their same-origin endpoint.\n\nIncludes an httptest regression covering JSON request forwarding and NDJSON response/status/header propagation.\n\nTargeted test: `go test ./scm -run TestHTTP`.


carli2:
denied

PR #819, opened 2026-09-07
Lower recursive planner folds in the optimizer

carli2:
## What

- restore `query_expr_alias_set` as the functional recursive `match`/`reduce` rule
- recognize the complete tagged-tree assoc-fold shape in `scm/optimizer.go`
- lower only an exact structural match to the stack-buffered physical tree walker
- keep that physical primitive optimizer-only: it cannot be called from Scheme source, does not appear in help, and is not documented
- leave partial or semantically different recursive folds untouched

This replaces the rejected direct planner call with an optimizer decision. It also corrects the stacked implementation that reached `master` through #825.

## Manual A/B measurement

Current `master` (`967538ea7`) versus this PR, normal Go build, fresh process and data directory per variant, identical 138-independent-subquery query from `tests/performance/traced-planner-scaling.yaml`, 5 warmups and 30 cache-busted `EXPLAIN COMPILE` samples:

| Metric (p50) | master | PR | Change |
|---|---:|---:|---:|
| HTTP wall | 572.670 ms | 568.911 ms | **-3.759 ms (-0.7%)** |
| compile total | 551.764 ms | 547.510 ms | **-4.254 ms (-0.8%)** |
| planner total | 245.113 ms | 241.511 ms | **-3.602 ms (-1.5%)** |
| untangle | 100.338 ms | 97.686 ms | **-2.652 ms (-2.6%)** |
| recipe emit | 142.384 ms | 139.412 ms | **-2.972 ms (-2.1%)** |

The parser phase was noise-positive (298.642 ms → 301.480 ms); it is outside this rewrite. Reorder was effectively neutral (2.554 ms → 2.643 ms).

## Verification

- optimizer exact-match and rejection tests
- optimizer-only source/help/documentation boundary test
- Scheme startup suite: 1280/1280
- SPARQL algebra: 21/21, including both MINUS cases
- decorrelated join reordering: 8/8
- compound ACL planning: 35/35
- Scheme lint for touched files
- normal build and `git diff --check`

The full SQL and JIT suites are left to CI as requested.

### DIFF
--- a/lib/queryplan-logical.scm
+++ b/lib/queryplan-logical.scm
@@ -922,10 +922,15 @@ the shallow guard instead of walking the wide expression it is about to drop. */
 projection for every source; that turns read-model queries into O(N^2) planner
 work before decorrelation has even started. */
 (define query_expr_alias_set (lambda (default_alias expr aliases)
-	(reduce (tree_collect_tagged_nth_unique expr (quote get_column) 1)
-		(lambda (found tblvar)
-			(set_assoc found (resolve_column_alias tblvar default_alias) true))
-		aliases)))
+	(match expr
+		((symbol get_column) tblvar _ _ _)
+		(set_assoc aliases (resolve_column_alias tblvar default_alias) true)
+		((quote get_column) tblvar _ _ _)
+		(set_assoc aliases (resolve_column_alias tblvar default_alias) true)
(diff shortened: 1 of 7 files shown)


carli2:
Closing as superseded by merged #823, which contains this optimizer lowering and its measured performance improvement together with the follow-up structural predicate recognition.

PR #823, opened 2026-09-08
Recognize recursive planner predicates in the optimizer

carli2:
## What

- restore alias-reference checks as ordinary functional Scheme recursion
- recognize the complete recursive `match` / short-circuiting `reduce` shape in `scm/optimizer.go`
- lower an exact match to one internal operand-tree predicate; structurally different functions remain untouched
- recognize the outer candidate reduction and consolidate repeated tree walks into one candidate-set walk
- remove the public planner-specific collector and predicate builtins introduced by the rejected stack
- prevent a recursive redefinition from inlining or specializing the previous same-named procedure
- keep internal physical primitives optimizer-only and unavailable to imported Scheme, help, and generated documentation

This directly addresses the review feedback: Scheme expresses the functional rule, while the optimizer detects and lowers the complete shape. Planner source no longer calls optimizer-only fused functions manually.

## Manual A/B measurement

Compared with stacked prerequisite #819 (`0b4f76726`), normal Go builds, fresh process and data directory per variant, the identical 138-dependent-subquery case from `tests/performance/traced-planner-scaling.yaml`, 5 warmups and 30 cache-busted `EXPLAIN COMPILE` samples:

| Metric (p50) | #819 | #823 | Change |
|---|---:|---:|---:|
| HTTP wall | 567.716 ms | 567.638 ms | -0.078 ms (-0.01%) |
| compile total | 547.134 ms | 547.344 ms | +0.210 ms (+0.04%) |
| planner total | 243.630 ms | 243.968 ms | +0.338 ms (+0.14%) |
| physical prepare | 2.020 ms | 1.999 ms | -0.022 ms (-1.08%) |
| recipe emit | 142.233 ms | 142.172 ms | -0.061 ms (-0.04%) |

The result is neutral within measurement noise. An intermediate implementation that materialized alias lists measured +1.7% HTTP and +3.3% planner; that experiment was removed before this push.

## Verification

- exact-match and near-miss optimizer unit tests
- candidate-set consolidation semantics, including excluding operator heads
- recursive-redefinition growth and result tests
- SPARQL graph algebra: 8/8
- complete SPARQL 1.1 algebra: 21/21, including cold and warm MINUS
- nested ACL callable cache: 7/7
- decorrelated join reordering: 8/8
- planner scaling: 8/8
- Scheme lint, Go formatting, normal build, and `git diff --check`

The full SQL and JIT suites are left to CI as requested.

### DIFF
--- a/lib/queryplan-logical.scm
+++ b/lib/queryplan-logical.scm
@@ -897,10 +897,6 @@ the shallow guard instead of walking the wide expression it is about to drop. */
 	(filter (coalesceNil sources '()) (lambda (src)
 		(and (not (source_outer? src)) (source_is_base_table? src))))))
 
-(define expr_referenced_aliases (lambda (default_alias expr)
-	(map (expr_collect_tagged_nth_unique expr (quote get_column) 1)
-		(lambda (tblvar) (resolve_column_alias tblvar default_alias)))))
-
 (define expr_refs_alias? (lambda (default_alias alias expr)
 	(match expr
 		((symbol get_column) tblvar _ _ _) (equal?? (resolve_column_alias tblvar default_alias) alias)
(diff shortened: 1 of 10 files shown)


carli2:
This patch manually inserts optimizer-only fused functions instead of improving the detection patterns in scm/optimizer.go

PR #824, opened 2026-09-08
Fuse profitable JIT map-reduce buffer loops

carli2:
## Summary

- fuse the accumulator loop over the existing row-major `mainBulkValues` buffer
- support arbitrary reducer arity and pass every storage column's exact `JITValueType()` into the inlined arithmetic
- keep the accumulator as a loop-carried, GC-safe phi home, preserving dynamic `nil -> int` and `int -> float` transitions
- use the calibrated fast path for simple reducers; other expressions use expression-dependent compile cost plus three real buffer A/B probes
- probe only when compile + probe costs at most 1% of the scan; retain fusion only above 2% measured benefit
- keep delta rows on the existing compiled-lambda path
- clear/pool the value buffer and drop query-local kernel/type references in `ShardMapReducer.Close()`

## Manual A/B measurements

AMD Ryzen 9 7900X3D, linux/amd64, patched Go 1.27 `GOEXPERIMENT=jit`, identical fixtures and 1024-row batches.

| reducer / rows | baseline median | best-of median | change |
|---|---:|---:|---:|
| SUM / 1 | 466.7 ns | 444.1 ns | -4.8% |
| SUM / 8 | 595.9 ns | 551.8 ns | -7.4% |
| SUM / 1,024 | 14.591 us | 14.646 us | +0.4% (no compile) |
| SUM / 60,000 | 819.0 us | 545.8 us | -33.4% |
| SUM / 1,000,000 | 14.593 ms | 7.735 ms | -47.0% |
| COUNT lambda / 60,000 | 720.0 us | 489.6 us | -32.0% |
| COUNT lambda / 1,000,000 | 12.083 ms | 6.871 ms | -43.1% |
| affine SUM / 1,000,000 | 20.545 ms | 20.564 ms | +0.1% (baseline retained) |
| two-column SUM / 1,000,000 | 22.603 ms | 22.565 ms | -0.2% (baseline retained) |

Complex control `(lambda (acc a b c) (+ acc (+ a (* b c))))`, corresponding to integer `SUM(a+b*c)`:

- current loop, 1,000,000 rows: 28.12 ms median
- forced fusion: 30.98 ms median, **+10.2%**, so it must not be selected on this machine
- best-of at 1,000,000 rows: no compilation; identical baseline path and 14 allocations/op
- simple kernel compile: 6.7 us, 7.9 KiB, 58 allocations
- complex kernel compile: 413 us, 1.21 MiB, 497 allocations

The calibration measures both compile anchors. At one million rows the complex query avoids compilation entirely. A sufficiently larger scan may run three probes; because forced fusion is slower here, it is rejected and remaining buffers use the compiled callback.

End-to-end `group-lookup-performance.yaml` cold ungrouped SQL SUM:

- master: 1,000,000 rows in 13.6 ms = 13.6 ns/row
- PR: 3,000,000 rows in 21.0 ms = 7.0 ns/row
- normalized improvement: -48.5%

## Validation

- normal and patched-JIT `go test ./scm ./storage`
- arbitrary three-column expression compilation and correctness
- point through million-row COUNT, SUM, affine, two-column and `a+b*c` spectrum
- dynamic accumulator transitions and deterministic query-local buffer release
- full CI matrix re-running for follow-up commit `2d4b3d84c`

### DIFF
--- a/scm/jit.go
+++ b/scm/jit.go
@@ -717,6 +717,42 @@ type JITStorageGetValueRangeFunc func(uint32, uint32, []Scmer, int)
 // JITStorageGetValueMultiFunc is the native arbitrary-record column-reader ABI.
 type JITStorageGetValueMultiFunc func([]uint32, []Scmer, int)
 
+// JITMapReduceBufferFunc folds rows from a row-major Scmer buffer. The row
+// width and callback are compile-time properties of the function; rows remains
+// explicit so zero-column reducers such as COUNT(*) need no synthetic values.
+type JITMapReduceBufferFunc func(Scmer, []Scmer, int) Scmer
+
+// PrepareJITMapReduceBufferProc returns source suitable for typed buffer-loop
+// inlining. Native declarations with a JIT emitter are wrapped in a procedure
+// so physical operators can specialize the same implementation without
+// teaching the storage package about individual aggregate names.
+func PrepareJITMapReduceBufferProc(source Scmer, arity int) *Proc {
+	if arity < 1 {
+		return nil
+	}
+	if source.GetTag() == tagProc && source.Proc() != nil {
+		return source.Proc()
+	}
+	if source.GetTag() == tagFunc {
+		if proc := JITProcForFunction(source.Func()); proc != nil {
+			return proc
+		}
+	}
+	declaration := DeclarationForValue(source)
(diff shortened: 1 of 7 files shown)


carli2:
converted to draft because:

- works for only one column
- certain other unclarities

PR #831, opened 2026-09-08
Replace serial callback adapters with reusable call frames

carli2:
## What and why

Remove `OptimizeProcToSerialFunction` completely, including its callers in list/association operators, storage, JIT preparation, the generator, and existing tests. There are no remaining references to the old API.

The common interface is now `PrepareSerialProc` + `Call([]Scmer)`. Serial operator loops own and reuse their argument arrays; parallel workers prepare independent programs. Constant results, argument projections, exact native forwarding, bound native operands and JIT procedures keep their specialized dispatch. Residual interpreted bodies use the same frame-ownership contract.

This is an execution-interface refactor, not a planner rule, cost-weight adjustment or JIT disablement.

## Correctness and ownership

- A callback may return its argument list or a closure. Declared retaining natives and unknown native callbacks receive an owning argument copy; variadic interpreted callbacks retain independent lists. Borrowing must not silently change observable results.
- Found and fixed an existing escaping-environment bug: a mapped closure could observe the parameter from a later callback invocation. Capturing interpreter bodies now receive invocation-owned environments. The new YAML closure test fails on the baseline and passes on this branch.
- Computed index readers prepare their callback and input buffer once per serial reader instead of allocating row argument arrays repeatedly.
- Native ordering functions retain their identity, which index/collation metadata depends on. The existing index-order Go callback ABI has an explicit adapter; this PR does not claim that arbitrary interpreted comparators are entirely allocation-free.
- JIT generation recognizes address-taken SerialProc receivers as callable descriptors, not heap-allocated Go structs. Their descriptors survive loop backedges, and known lambda templates still inline. The dynamic JIT callable-value ABI retains a small private adapter; ordinary operator loops do not use that variadic wrapper.

The large diff in `scm/list.go` and `scm/list_assoc_extra.go` is predominantly regenerated emitters. Review their `Fn` bodies together with `tools/jitgen/main.go`; generated emitters follow the new reusable-array source pattern. No generated cost weights were hand-tuned.

## Manual A/B measurements

Baseline: current master `7db1a4dd3` (#829). The measurement baseline was its tree-identical pre-merge commit `85d43b2ab`. Candidate: this API migration. Same machine (Ryzen 9 7900X3D), compiler and fixtures for each A/B pair. These measurements were collected before opening this PR; CI is additional validation.

### Native operator / interpreted callback microbenchmark

Map 1,024 integers with a prepared raw interpreter callback, invoking the native map implementation to avoid optimizing away the benchmark itself. Ordinary Go build, `-benchmem -benchtime=300ms -count=3`; sequential baseline then candidate, median of three samples. The temporary benchmark harness is not added to the permanent Go test suite.

| Callback | Master | Branch | Change | Allocations, master → branch |
|---|---:|---:|---:|---:|
| identity | 29.558 µs | 5.623 µs | −81.0% | 1,027 → 2 |
| add constant | 83.234 µs | 13.707 µs | −83.5% | 1,037 → 3 |
| compare with constant | 78.463 µs | 12.562 µs | −84.0% | 1,037 → 3 |

Bytes per operation: identity 34,848 → 18,448; arithmetic/comparison 35,464 → 18,480. Output allocation remains necessary. The bound-native two-argument frame is allocated once at preparation, not once per element.

### Complete query compilation, not just the microbenchmark

Two offline copies of the same application fixture, same SQL/bindings and JIT build. `EXPLAIN COMPILE` through the existing `tools/benchmark_compile.py` request implementation; 3 warmups then 60 measured samples per side, alternating AB/BA and using unique SQL comments. No production instance was changed.

| Query shape | Master p50 | Branch p50 | Change | Master p95 | Branch p95 |
|---|---:|---:|---:|---:|---:|
| 28.8 KB search/ACL list, LIMIT 72 | 206.899 ms | 201.975 ms | −2.4% | 253.974 ms | 254.417 ms |
| Related ACL count | 130.610 ms | 131.987 ms | +1.1% | 162.275 ms | 182.956 ms |

Plan sizes and structural metrics match: 57,328 bytes for the list and 12,410 for count. This is a modest list compile improvement, **not** a claim of faster compilation for every query or lower application request latency. Count is slightly slower and has a worse tail in this run. A preceding 20-sample count run showed +10.5%, its repeat +0.8%; the expanded sample records that variability rather than hiding it. The shared development host is noisy. CI A/B remains enabled and must pass.

## Validation

- JIT and ordinary Go builds succeed.
- Startup Scheme unit tests: 1,276/1,276 with the final JIT build.
- New `tests/sql/expressions/serial-callback-frames.yaml`: 9/9 with JIT and 9/9 without it. Covers retaining natives, variadics, escaping closures, dynamic compiled callbacks, reducers, bound operands, parallel workers, empty inputs and error propagation.
- Existing native ORDER/LIMIT YAML: 12/12.
- Existing RecSet operator YAML: 34/34.
- Existing trace-derived planner scaling YAML: 8/8; its optional PERF_TEST case was not run locally and is not counted as a pass.
- Existing targeted SerialProc Go tests and compiled-order-access test pass. Their API call sites were migrated without weakening assertions; the native-forwarding equivalence check now uses independent `Apply` semantics as its reference.
- Full suite and performance gates belong to CI, per project instruction. No full local suite, disabled test or relaxed performance gate.

No persistent format, data-directory lifecycle, index selection heuristic, or live-server configuration changes.


carli2:
Fixed the CI allocation regression in 3823c9b89. Both go-test and jit-go-test failed TestCompiledScanLookupCompositeAllocationLimits/map: 6 allocations instead of the unchanged maximum of 5. Reproduced locally before the fix. The lookup already owns a freshly allocated result array, so CallOwned now transfers that array to retaining native callbacks without a redundant defensive copy. Reusable loop frames still use Call and retain its ownership protection. The original allocation assertion passes with ordinary Go and the JIT compiler; targeted lookup and SerialProc tests also pass. No test thresholds or CI mechanisms changed.

carli2:
Cherry-picked e60f5fc5f and b171525f3 as requested (branch commits 1e7f50542 and c7ba68209), without conflicts. Targeted validation on the combined build: RDF algebra 21/21, including both previously failing MINUS cases; index-delta-merge 43/43, including Boolean/text and repeated rebuild coverage. Targeted binary collation, compression identity and existing lookup allocation tests also pass. Full suite remains in CI.

PR #838, opened 2026-09-08
fix(storage): recover index-build panics and protect shared blobs

carli2:
## Summary

Fix two independently reproduced storage failures:

* A column read can panic while an auto-index is built. The index mutex previously remained locked, so later scans waited forever. Release it on unwinding and discard partial build state; preserve the original error and permit a clean retry.
* Operational blob refcounts can undercount committed owners. Dropping one owner previously deleted a shared payload when the counter reached zero. Decrements now retire only the counter row. Physical reclamation belongs exclusively to the existing committed-generation cleanup proof.

No planner/cost-model changes, no suppression of missing-blob errors, and no return-NULL fallback.

## Regression evidence

New permanent YAML tests, written before their respective fixes:

* `index-build-panic-recovery.yaml`: on the master binary the retry times out after 10 seconds; fixed binary passes repeated injected failures followed by correct successful reads. The error is injected through a computed-column reader used by the ordinary ordered auto-index path, not a new production fault hook.
* `blob-undercount-recovery.yaml`: three externally stored payloads are shared by two tables. After deliberately undercounting only those hashes and dropping one table, the unfixed implementation raises `OverlayBlob: missing blob`. The fixed implementation retains all three exact values.

Both suites pass with the ordinary and JIT builds. Existing warm-index-hook consumer YAML: 8/8. Targeted existing blob lifecycle and cleanup Go tests pass.

The existing file-reclamation tests retain their exact assertions; they now explicitly invoke `CleanDatabase` after owner retirement, reflecting the intentional contract that a counter decrement is not deletion authority. Reclamation is deferred, not disabled. This may retain orphaned payloads until cleanup runs.

Full SQL/JIT/performance suites are delegated to CI; no local full-suite run and no relaxed test expectations or timing thresholds.

## Incident attribution

The lock leak and undercount data-loss mechanism are reproducible. They do not prove which historical operation removed a particular payload from an external deployment. Missing payloads must be restored from hash-verified copies separately; this patch cannot recreate lost contents.

### DIFF
--- a/storage/blob-refcount.go
+++ b/storage/blob-refcount.go
@@ -177,7 +177,13 @@ func (db *database) IncrBlobRefcount(hash string) {
 }
 
 // DecrBlobRefcount decrements the reference count for a blob hash in db.`.blobs`.
-// If the count reaches 0, the row is deleted and the blob file is removed.
+// If the count reaches 0, only the operational counter row is deleted.
+// A counter can undercount committed owners after a crash or an interrupted
+// lifecycle update. It is therefore NEVER authority to delete payload data.
+// CleanDatabase reclaims unowned files using the complete committed-generation
+// manifests under persistenceLifecycle. Do not scan those manifests here:
+// callers can already hold schema/shard lifecycle locks, and doing so would
+// introduce lock inversions as well as a full catalog walk per decrement.
 func (db *database) DecrBlobRefcount(hash string) {
 	defer db.lockBlobRef(hash)()
 	state := db.blobRefState()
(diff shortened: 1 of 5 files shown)


carli2:
All correctness jobs are now green, including the full SQL and JIT suites, Go/JIT-Go, alternate storage and packaging.

The first performance run failed only the first-fill wide arithmetic aggregate: 62.749 ms baseline versus 198.256 ms candidate, one timed sample. I have not changed its threshold, sample count, fixture or the CI gate.

Targeted manual investigation with ordinary Go builds, the same 60,000-row YAML setup recreated for every measurement, six first-fill samples per version, alternating A/B and B/A:

* master: 39.560, 27.071, 26.560, 25.290, 26.609, 26.591 ms; median 26.600 ms
* candidate: 28.476, 27.575, 26.408, 28.549, 28.642, 25.700 ms; median 28.025 ms (+5.4%)

This does not reproduce the threefold CI slowdown and does not prove zero performance impact. The failed performance job has been retried unchanged; its result remains required.

carli2:
All checks are now green, including performance A/B on the unchanged retry (run 34273502525, job 102227926191). No code, baseline, timing threshold or CI gate was changed for the retry. The ordinary and JIT correctness suites passed independently.

PR #872, opened 2026-09-09
Fix inclusive RAM accounting and cascading eviction ownership

carli2:
## Problem

A shard could report less RAM than one of its own LIKE indexes. The display used an exclusive eviction-registration size and table rows used a stale planner statistics snapshot. Resident temporary columns, decoded dictionaries and blob caches were not consistently represented in parent totals.

## Ownership contract

- `ComputeSize()` is inclusive: object layout and owned backing allocations/children. Separately registered children are subtracted explicitly for the global eviction ledger, not hidden inside the public size contract.
- Structured `show` shard/table sizes and the dashboard include resident indexes (including inverse positions/hooks), temporary columns, dictionaries and decoded blob caches. Cold columns are not loaded to inspect their size.
- Fixed layouts and backing capacities replace several stale constants/length-only counts. Embedded storage structs are not counted twice; decimal file-format bytes are not RAM.
- Full parent eviction offers include separately registered children. Busy indexes/nested caches stay registered; successful child releases are accounted once even if a later child prevents parent eviction. Wrapper traversal in the manager is nonblocking. Dictionary eviction no longer counts a release twice.
- Per-shard temporary-column portions are published in mutation order, transferred to replacement generations and subtracted only when their buffers are detached. Unchanged sizes do not send another manager operation.
- Existing eviction type weights and savings policy are unchanged. Bytes and replacement cost remain distinct.

The contract and the prohibition on full ownership walks in hot paths are documented in AGENTS.md. Full traversal is limited to explicit diagnostics/size publication, not query rows, cost guards or planner-statistics reads. `stat` exposes process RSS and cached Go heap/stack diagnostics as separate views.

## Scope and precision

This fixes resident storage ownership and cascading eviction accounting; it does not equate database payload with process RSS. Go map/bucket overhead and some allocator/native-runtime bookkeeping still use estimates. Runtime code, stacks, allocator slack, transient query/rebuild buffers and shared process allocations cannot be assigned to a table by summing storage payloads. Cached runtime samples may be older than the RSS reading. No rescaling or fabricated per-database attribution is used.

A separate disposable 8,000-row diagnostic run reported the main table as 522.5 KiB base + 265.8 KiB indexes + 1.635 KiB metadata = 789.9 KiB. The database owner estimate was 804.4 KiB. Process RSS was 137,089,024 bytes and the global eviction ledger 2,833,590 bytes: these are explicitly different measurements, not proof of exact heap reconciliation.

No live data, persistence format, retention setting or planner/calibration weight was changed.

## Test-first evidence

New isolated `tests/storage/persistence/shard-memory-totals.yaml` covers index-heavy shards, temporary columns, live table totals, disjoint categories, blob RAM, invalid shard access, a 256-row small-shard fixture, recursive eviction, cold diagnostics without reload, and exact persisted data after reload.

The initial three accounting assertions failed on the unchanged baseline before implementation. Re-running the expanded suite against current master `a056e7147` also fails the index/temp/table/blob/category assertions; the proposed code passes **22/22**.

Targeted validation (no local full suite):

- New ownership YAML: **22/22**.
- Existing memory-pressure YAML: **32/32**.
- Existing memory-engine YAML: **24/24**.
- Existing cache-engine YAML: **24/24**.
- Targeted Go ownership, index, RecSet/inverse traversal and eviction tests pass.
- Targeted race-enabled ownership/eviction tests pass.
- Build, changed Scheme lint and diff checks pass.

Known baseline issue: both unchanged master and this branch print startup unit failure 206, `membership text guards use the prior without emitting a zero-hit scan` (1275/1276). This unrelated assertion has not been disabled or modified. Full functional/JIT/performance suites belong to CI.

## Manual A/B hot-path check

Same machine (Ryzen 9 7900X3D), Go toolchain, existing benchmark fixtures, 200 ms/sample, three samples each; medians below. Baseline is `a056e7147`. These are microbenchmarks, not application end-to-end claims.

| Existing workload | Master | Proposed | Change |
| --- | ---: | ---: | ---: |
| Published table statistics | 0.769 ns | 0.854 ns | +0.085 ns |
| Planner statistics hash lookup | 30.78 ns | 31.85 ns | +1.07 ns |
| Scan COUNT, 60,000 rows | 132.10 us | 132.39 us | +0.22% |
| Filtered COUNT | 1.864 us | 2.045 us | +9.7% |
| Scan SUM | 580.76 us | 545.13 us | -6.1% |
| Scan take-right reducer | 136.82 us | 122.57 us | -10.4% |

Allocations/op are unchanged in every listed workload (statistics reads remain allocation-free). Short-sample variation is visible, so this is a check against accidentally adding a memory walk to scans, not a speedup claim; CI A/B remains enabled.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -188,6 +188,31 @@ curl -s -u root:admin "http://localhost:[PORT]/sql/DBNAME" -d "SELECT 1"
 - Join pipelines: drive ordered/filtered side; hash/range-probe the other; keep fuse-friendly structure; precompute hidden computed columns for FK/PK group reuse.
 
 ## Memory & CPU Efficiency
+
+### Inclusive size and eviction ownership contract (2026)
+
+- `ComputeSize()` includes the object's retained allocations and all owned
+  children. It is never an eviction weight or an exclusive registration size.
+- A global ledger charges each allocation once. Where children have their own
+  registrations, subtract them explicitly at the registration boundary.
+- Dashboard/show resident totals include indexes, temporary columns, decoded
+  dictionaries and blob caches, whether evictable or not. Do not substitute
+  cached planner sizes or the eviction ledger for these totals; never load cold
+  columns just to measure resident memory.
+- A parent eviction offer includes all children released by that action. Parent
+  and child offers can overlap; actual releases are booked exactly once, only
+  after successful cleanup. A failed TryLock cannot authorize deregistration.
+- Memory bytes, reclaimable bytes and replacement-cost weights are distinct.
+  Changing accounting does not authorize changing the policy weights.
+- `storageShard.tempColumnBytes` is protected by the shard mutex. It records
+  per-shard portions of a separately registered temporary column so the manager
+  can offer/release those portions without blocking on column computation.
+- Resident payload estimates and process RSS are not interchangeable. Keep
+  allocator/runtime/stack and unassigned process memory visible; do not force
+  agreement by rescaling database sizes.
+- Full ownership traversal is an explicit diagnostic/maintenance operation,
+  not a query, scan-row, guard, or planner-statistics hot path. Reuse published
+  scalar statistics and incremental cache-ledger updates in those paths.
 - Design principle: Cache misses are more expensive than lightweight compression. Prefer compact encodings (e.g., bit-packing 3/5‑bit integers) and sequential scans over scattered, cache‑cold access.
 - Use columnar storage to keep footprints small and hot; compress where it reduces cache lines touched even if it adds tiny (de)compression overhead.
 - Pull function calls out of loops whenever possible
--- a/lib/dashboard.scm
+++ b/lib/dashboard.scm
@@ -120,9 +120,10 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 	)
 ))
 
-/* helper: sum size_bytes across all shards of a table */
+/* Resident table total includes metadata and all child allocations. Never use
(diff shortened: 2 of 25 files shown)


carli2:
CI follow-up: normal SQL, Go/JIT-Go, packaging, alternative storage and both performance A/B jobs passed. The first JIT SQL run timed out on two concurrent blob readers under memory pressure. A targeted local JIT run of blob-storage.yaml passed 79/79; an additional disposable stress run passed 600 checked blob reads alongside 100 RAM-budget pressure cycles. No test, timeout or safety gate was changed. The failed JIT job has been rerun once and is being monitored; the original timeout is not yet reproduced or explained.

carli2:
The retry passed, but pressure testing reproduced a real lock cycle: CacheManager cleanup entered dropTrigger and waited for an unrelated table DDL lock, while that table rebuild waited for CacheManager.Remove. Commit 9715a66f7 resolves trigger ownership before taking the DDL lock and revalidates under the existing locks. No lifecycle callbacks or JIT paths were disabled. The deterministic lock-order test failed before the fix and passes afterwards (also under the race detector). The extended isolated YAML suite passes 26/26; a separate JIT pressure run passed 1,000 checked blob reads plus 200 distinct COUNT compilations. No test gates or time limits changed. Monitoring the new CI run.

carli2:
Correction to the timing of the previous update: #872 had already been merged at 1c7854702 before the follow-up push. All checks on that merged commit are green. The subsequently discovered and tested trigger-cleanup lock fix is now a separate PR, #875, based on current master.

PR #875, opened 2026-09-09
Prevent eviction deadlocks in trigger cleanup and memory diagnostics

carli2:
## Evidence and fixes
Three proven lock cycles:
1. Trigger cleanup locked unrelated table DDL before checking ownership. A group-table rebuild could hold that lock while waiting for CacheManager.Remove. Resolve ownership first and preserve revalidation.
2. CI goroutine dumps show CacheManager waiting in dropTrigger for the schema lock; SHOW holds that lock while waiting for blobRAMCache.cachedBytes; a blob reader holds the blob mutex while waiting for CacheManager.Stat. Publish resident bytes atomically at batch/eviction boundaries so size diagnostics never acquire the reader mutex.
3. Rebuild held the successor mutex, published nextReady, then reacquired the source mutex to transfer RAM accounting. Transaction commit takes source -> successor. Transfer accounting inside the existing final source critical section, before enabling forwarding, in both rebuild branches.

ComputeSize remains inclusive. No memory traversal enters scan hot paths. Lifecycle callbacks, maintenance triggers, JIT and eviction policy stay enabled.

## Verification
- All three deterministic lock regressions fail before their respective fixes and pass afterwards. Targeted race tests pass, including three repeated runs of the final three-test selection.
- Extended isolated RAM YAML suite: 29/29 on the current head. Its new parallel UPDATE/rebuild case verifies the committed payload afterwards.
- Blob YAML adds concurrent SHOW snapshots; all original assertions remain intact. Prior head passed 80/80 locally with JIT and in JIT CI.
- Manual anonymous-fixture stress passed 10,000 checked blob reads, 1,000 budget cycles, 1,000 COUNT compilations, and 100 rebuild/pressure cycles. Those did not expose the SHOW cycle; CI stacks did.
- Repeated parallel runs of four unchanged aggregate suites exposed a separate stale-SUM bug. It also reproduces on 43d0c8140, before #872. Tracked independently in #880 with reproduction details; it is not claimed fixed here.

Temporary diagnostic commits remain reviewable in history, but the final diff removes their helper and restores both CI workflows byte-for-byte to master. No test limits, gates or existing assertions weakened. Full suites and A/B only in CI. Final head 648c874f2: all applicable CI checks passed. SQL 14m27s; JIT SQL 20m33s; Go 2m02s; JIT Go 5m49s; A/B 18m30s; JIT A/B 17m20s; alternative storage, packaging and policy guards passed. PHP was skipped by the existing workflow. JIT CI explicitly confirms RAM 29/29 and blob 80/80 with zero failures in those suites. Existing noncritical failures elsewhere remain reported by the unchanged runner; green CI is not a claim that every noncritical case passed.


carli2:
Not merge-ready yet: JIT SQL still times out in blob_ram_pressure, despite the deterministic DDL-cycle test and local JIT suites passing. Both performance A/B checks pass. I have marked this PR draft and added temporary opt-in CI stack diagnostics (no cancellation, timeout increase, skipped assertions, or changed exit-status semantics). This diagnostic code will be removed once the remaining stalled stack is captured and addressed.

carli2:
CI stack evidence identified the remaining cycle precisely: CacheManager -> schema lock (dropTrigger line 523), SHOW -> blob mutex (cachedBytes), blob reader -> CacheManager.Stat. Commit adb9b44e9 publishes the blob RAM footprint atomically at batch/eviction boundaries. The new deterministic size-lock test fails before the fix and passes after it; targeted race tests pass. Final local JIT results: blob suite 80/80, RAM suite 26/26. Temporary diagnostics are deleted and both workflow files match master exactly. Monitoring final CI now.

carli2:
Additional reproduced CI blocker fixed in 648c874f2: rebuild held the successor mutex, published nextReady, then reacquired the source mutex solely to transfer temporary-column RAM accounting. Commit takes source -> successor, creating a lock cycle. The transfer now happens inside the existing final source critical section before nextReady in both rebuild branches. A deterministic test fails on adb9b44e9 and passes after the fix (also with -race, three repetitions). The isolated RAM YAML suite now passes 29/29, including concurrent UPDATE/rebuild and payload verification. Existing aggregate tests remain unchanged. Parallel repetitions also exposed an intermittent stale SUM after UPDATE; investigation continues, so this PR remains draft. New CI is running; previous JIT SQL and both A/B checks were green, but that is not a substitute for validating the new head.

carli2:
Final head 648c874f2 has completed all applicable CI checks successfully, including both SQL jobs, both Go jobs, both performance A/B jobs, alternative storage and packaging. RAM 29/29 and blob 80/80 are explicitly green in JIT CI. No retries were needed for this head. Marked ready for review; not merged. Separate pre-existing stale aggregate race remains tracked in #880.

PR #885, opened 2026-09-10
Perf-test KPI plan-cache reuse across rotating period bounds

carli2:
Stacked on #883 (carrying the #884 planner fix as its second commit); targeted at master. Until #883/#884 land, this PR shows their three commits and then slims to the one new perf suite.

## What this measures

`tests/performance/kpi-period-cache-reuse.yaml` reproduces the generated-dashboard workload behind #883: a KPI query whose period derived table is `SELECT <const> AS period, <const> AS s, <now()-varying const> AS e UNION ...`, so the query text and the plan-cache key change on every request.

Two cases, each 20 executions through one shared `newcachemap` with a rotating period-0 bound:

| case | shape | guards |
|---|---|---|
| KPI plan is reused across rotating period-bound literals | `COUNT(1)` + nested `SUM((SELECT SUM ...))` | #883 constant-projection-row parameterization |
| KPI COUNT(DISTINCT) keeps every period and reuses the plan | + correlated `COUNT(DISTINCT c.customer)`, asserts the empty trailing period still yields a row with `0` | #884 empty-group row-drop fix |

## Results (local, `PERF_TEST=1`)

- Branch: 2/2 green, ~1.2s / ~0.6s vs a 4s threshold, stable across repeated runs.
- master: warmup fails outright — the churn of near-identical constant-folded plans through the cache trips a non-deterministic `bind_query_names: unknown relation alias: h` on the empty trailing period. #883's parameterization keeps the derived-table columns non-constant, removing the re-plan cost and sidestepping that bind path. (#884's logical-plan diff is a no-op for case 1; case 1 is carried entirely by #883.)

`performance-ab` / `jit-performance-ab` run the suite against both the PR-merge binary and the base (`master`) binary; the base side is expected to report the warmup failure above until #883 merges.

## Notes

- `metadata: ci:false, perf_ci:true` — picked up by the performance-CI batch only, not the SQL correctness run.
- Fixture: 5 non-empty periods × 1000 offers + 5000 line items (`ENGINE=sloppy`), built once in case 1's per-case `setup`; case 2 reuses it, matching `kpi-period-aggregates.yaml`.

https://claude.ai/code/session_01RSuVhbwv6pvf3WiNPCRUht

### DIFF
--- a/lib/queryplan-logical.scm
+++ b/lib/queryplan-logical.scm
@@ -3087,6 +3087,14 @@ without separately proving two-valued semantics. */
 			where_corr_pairs having_corr_pairs))))
 		(define local_value_expr (decorrelate_expr_with_pairs inner_default all_corr_pairs value_expr))
 		(define ags (dedupe_aggregates_by_col (merge (list (extract_aggregates local_value_expr) (list aggregate_count_descriptor)))))
+		/* COUNT(DISTINCT)/GROUP_CONCAT(DISTINCT) lower to a merge_unique LIST
+		accumulator whose empty-group row the physical group cache does not emit, so
+		`preserve_empty_domain` cannot make an INNER join safe here. Keep the stage
+		source as a real LEFT join; a no-match outer row then reads NULL and
+		count_distinct_read_expr / group_concat_distinct_read_expr coalesce it to
+		0 / NULL. Neumann NK15 3.2: outer joins keep unmatched tuples. */
+		(define ags_have_list_accumulator (reduce ags (lambda (found ag)
+			(or found (count_distinct_descriptor? ag))) false))
 		(if (empty_list? ags)
 			(neumann_fail "untangle_query" "table-backed scalar subquery without aggregate needs partition_limit=2 and overflow checking")
 			true)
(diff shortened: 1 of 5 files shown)


carli2:
Parking as draft — this depends on #883 (the projection-literal parameterizer), which is still being reworked. The scenario under test only exists once literals are parameterized into session vars; without #883 the planner constant-folds the period bounds and the subquery is never correlated.

Findings from investigation, to fold in when #883 stabilises:

1. **`bind_query_names: unknown relation alias: h` in CI was a test bug, not a planner bug.** `(string (+ 6000000 i))` renders as `"6e+06"` (scientific notation) in this Scheme; `SELECT 6e+06 AS e` breaks alias binding of the derived table. Fix: keep rotating bounds < 1e6 (e.g. `(+ 900000 i)`) so `(string …)` stays plain-integer.

2. **Better assertion than a wall-time threshold:** after 20 rotating-literal executions through one `cached_parse` cache, assert `(count cache) == 1`. Verified: master → 20 queryplan-cache entries, #883 → 1. Deterministic and fixture-independent. On the small fixture the wall-time delta is only ~2 ms, so a timing threshold guards nothing; the real multi-second effect needs the big `kpi-period-aggregates` fixture.

3. **#884's empty-group fix is confirmed independently** (`@w` session-var bounds, empty inner range: master returns 0 rows, #884 returns the row with `c = 0`). #884 does not need this PR.

PR #891, opened 2026-09-12
Replace the global JIT builtin-inline budget with local per-call admission

carli2:
Builtin inlining now uses the call's existing argument-type, shape, callback, and generated-body checks without an accumulated per-function budget or a live-register-count veto. Previously, unrelated earlier inlining or register allocation could change the decision for the same call arguments.

The representation guard rejects unmaterialized lambda templates on parameters that cannot handle them, and the generated-body sentinel remains enforced. No estimated call-boundary formula or replacement register threshold is introduced.

The branch is rebased onto master `9df0941f3`. The added admission test exercises identical call shapes with exhausted, available, and protected register sets, including cases that must remain rejected.

Validation with patched Go `84fe25ee45468f5e3920c11c24e2bbb926fc3a72`:
- Targeted JIT admission, compiler/callback, register, and ABI tests passed.
- `tests/sql/expressions/merge-singleton-append.yaml`: 5/5 passed, including `join_optimizer_reorder_sources` compilation and the invalid-merge case.
- Full-suite and performance validation run in CI, as requested. No local full-suite run was performed for this update, and no performance improvement is claimed from these correctness probes.


carli2:
**Review: the `FreeRegs` snapshot in `jitInlineRegisterHeadroom` is the wrong signal for this gate**

Replacing the accumulated `ctx.BuiltinInlineCost` budget is the right call — a serially-accumulated, AST-encounter-order-dependent unit sum was never a real resource, and the "two inert unused defines shift generated code elsewhere" repro nails exactly why.

But `jitInlineRegisterHeadroom(ctx) >= jitMinInlineRegisterHeadroom`, reading `ctx.FreeRegs &^ ctx.ProtectedRegs` live at each call site, reintroduces the same non-locality this PR set out to remove — just at a smaller scope, and with a real quantity standing in for a fake one instead of the problem going away:

- **Still order-dependent within one function.** `FreeRegs` at the moment call N is considered depends on exactly how many registers earlier calls in the *same function's* AST-encounter order have already claimed and not yet freed. Whether call N inlines can still flip based on unrelated, textually earlier code in the same function — the original bug, just no longer crossing function boundaries.
- **Conflates "registers are tight right now" with "this call needs registers."** A call whose arguments are compile-time constants (or become constant after upstream folding) can inline down to a single value with zero runtime register footprint — `FreeRegs` doesn't know that and refuses it exactly as readily as a call with fully dynamic, register-hungry arguments. Conversely, a call that genuinely does need live registers can still get admitted early, before `FreeRegs` has dropped, and only starve something inlined later in the same function.
- **Doesn't compare against the actual alternative.** The first version's `cost <= jitCallBoundaryCost(argc)` was closer in spirit: a call whose own inlined cost is no worse than the call boundary it would replace is safe to inline purely on structural grounds, independent of the rest of the function — a real, local, deterministic comparison. It needed the register-exhaustion case (`merge-singleton-append.yaml`'s `join_optimizer_reorder_sources`) and the `jit-performance-ab` regression fixed, but the fix that landed (a live register-count veto) discards that comparison's determinism rather than repairing it.

Suggested direction: keep a per-call-site *structural* comparison as the primary admission signal — this call's cost after accounting for its own folding/specialization potential (not just the flat declared `JITInlineCost`) against `jitCallBoundaryCost(argc)` for its actual argument shapes — and find a bound for the register-exhaustion case that doesn't depend on live allocator state at arbitrary AST-encounter order. A static worst-case register estimate per callee, or a genuine two-pass admission that ranks candidates by cost/benefit before committing any of them (rather than greedy left-to-right admission gated by a runtime snapshot), would keep the fix local and deterministic the way the rest of this PR's reasoning already argues for.

PR #908, opened 2026-09-14
Improve cost estimates for selective grouped joins; include TPC-H correctness fixes

carli2:
Selective six-table grouped joins can choose a tiny dimension followed by a repeatedly executed composite subtree while postponing the selective part filter. Integer compression also reports row counts as a loose distinct-value upper bound, exaggerating repeated foreign-key domains and distorting join cardinalities.

Bound integer distinct estimates by the already-published rebuild domain (including NULL), without reading shards or adding statistics traversals. Charge composite right-subtree execution once per left binding and charge unsuccessful join probes too; retain reusable build and compilation costs once. Both logical cost domains and generated runtime cost expressions use this model. There are no query-name special cases, alternate subquery fallbacks, new storage operators, or persistence format changes.

This PR targets **master** and now includes merged #907 through current master `39282d864` plus the rebased #909 head `dbebda2c4`. It can land with both correctness fixes without additional merges.

### Manual A/B measurements

Same machine (Ryzen 9 7900X3D, 24 logical CPUs), JIT enabled, GOMAXPROCS=24, HTTP SQL frontend. Compare master `508f999c4` against this branch, using the exact same `tests/performance/joined-aggregate-driver.yaml` setup: 8,192 fact rows, six tables, rebuild, two warmups, five timed samples, median latency. Fixture sizes are literal and identical on both sides.

| Automatically planned workload | Master | Proposed | Change |
| --- | ---: | ---: | ---: |
| Selective six-table profit grouping | 286.1 ms | 5.0 ms | -281.1 ms (-98.3%, 57.2x faster) |

An earlier independent run measured 282.6 ms versus 6.0 ms. Both sides returned the same independently computed 16 grouped sums; the empty-filter case also passes. These are manual measurements taken before opening this PR, rather than CI results.

Development inspected EXPLAIN, EXPLAIN IR, PHYSICAL and REORDER, extracted scan timings and alternative join orders/operators. On the existing SF 0.1 data, extracted green-part selection cost ~2.1 ms, partsupp scan ~0.9 ms, and lineitem scan ~2.2 ms; a selective-part-first hand plan was substantially better than the original plan. The changes make such choices available to the generic cost model. The complete SF 0.1 automatically planned remeasurement is now available below. No PostgreSQL win is claimed.

### Validation

- Targeted storage tests: `TestShowColumns*` and `TestPlannerStatistics*`, including repeated/nullable integer domains, sparse domains, overflow and decimal exclusions.
- Standard and JIT SQL: the complete ORDER/LIMIT suite passes 59/59, including an explicit synthetic check that a bushy join remains incompatible with the left-deep ordered pipeline. The cost-selected SQL plan uses `scan_join_order` without Scheme sort/slice materialization, and its following result oracle passes.
- Standard/JIT joined-aggregate needle suite passes 3/3. Warm medians on the merged head are 9.0 ms standard and 5.9 ms JIT.
- `git diff --check` passes.
- Full SQL/JIT suites and performance regression A/B are left to CI, per the requested workflow; committed with `--no-verify` after targeted checks.

The earlier Standard/JIT CI failure was an obsolete physical-plan assertion: the cost change legitimately selects `scan_join_order`, while the result oracle remains correct. The regression now tests bushy-tree rejection directly instead of inferring it from the chosen SQL plan.

The principal risk is changing generic join selection on other workloads. CI's performance A/B gate must confirm that no measured case regresses by more than 20% before merging.

### Follow-up measurement on PR head cced90cdf (existing SF 0.1, JIT)

| Query | PR908 median ms | PG median ms | PR908 / PG | First run ms | Correctness |
|---|---:|---:|---:|---:|---|
| Q9 | 515.65 | 82.69 | 6.24x | 953.76 | PASS, all 7 responses |
| Q19 | >60000 (first request timeout) | 3.17 | >18941x | >60000 | No result |
| Q18 | 12695.31 | 205.88 | 61.66x | 13297.46 | PASS, all 7 responses |
| Q13 | 7833.22 | 54.28 | 144.30x | 7968.84 | PASS, all 7 responses |
| Q21 | 794.04 | 44.47 | 17.85x | 884.09 | PASS, all 7 responses |
| Q8 | 65.06 | 27.81 | 2.34x | 333.17 | PASS, all 7 responses |

Same-machine PostgreSQL 12.22 results were reused from the earlier native-socket run; MemCP uses HTTP. Both use two warmups and five samples. Every completed response matches the same-SF oracle at canonical display precision. Q19 timed out on its first request after 60 s; its engine was stopped before continuing with a fresh fixture. No warm median exists for Q19. Next priority is Q19 shared predicate extraction from OR branches, followed by Q18 and Q13. Raw data and all EXPLAIN variants are retained locally under `tpch/runtime/pr908-remeasure/`.


carli2:

PR #910, opened 2026-09-15
perf(planner): expose disjunctive join predicates for TPC-H Q19

carli2:
TPC-H Q19 hides its join equality and shipping filters in three OR branches. The planner therefore selects nested scans and the SF 0.1 query times out after 60 seconds. Derive structurally identical shared conjuncts and necessary per-relation OR filters after name binding, retaining the complete original OR residual. This exposes the existing `scan_join_order` alternative to the current cost model, without query-specific rules or new operators.

Based on #908 at eecbb9b10, including the merged #907/#909 fixes and current master 68f8092b6. Target is master; until #908 merges, this PR also contains its commits.

### Manual measurements before opening this PR

Ryzen 9 7900X3D, 24 logical CPUs, GOMAXPROCS=24, JIT enabled and native procedure JIT verified. Same prepared SF 0.1 data, SQL, HTTP response decoding, two warmups and five measured samples. No data regeneration. Baseline is eecbb9b10, candidate is this branch.

| Query / fixture | Baseline median | Candidate median | Change |
|---|---:|---:|---:|
| Original Q19, 128 parts / 1024 lineitems regression fixture | 17.307 ms | 1.424 ms | -15.882 ms / -91.8% |
| Already manually factored Q19, same small fixture | 1.037 ms | 1.308 ms | +0.271 ms / +26.1% |
| Original Q19, SF 0.1 | First request timed out at 60000 ms; no warm median | 7.323 ms | Censored baseline; no finite warm speedup claimed |

SF 0.1 candidate measured samples: 7.820, 7.273, 8.017, 6.187, 7.323 ms. Both warmups and all five samples match the same-SF PostgreSQL oracle for every output at canonical TPC-H output precision. Same-machine PostgreSQL 12.22 warm median was 2.568 ms (native socket), so MemCP remains approximately 2.85x slower. This is warm execution; not a cold-cache win claim.

The additional local filter costs 0.271 ms on the already rewritten tiny fixture; retained explicitly above. CI A/B, including its absolute noise tolerance, remains unchanged.

### Plan / profile evidence

EXPLAIN, IR, REORDER and PHYSICAL were collected. Original CPU profile is dominated by string decoding, batched filtering and SQL equality inside the nested scans. Hand plans with existing operators established the target before implementation: legacy part-first 828.97 ms / lineitem-first 124.01 ms versus fused part-first 16.68 ms / lineitem-first 8.65 ms (cached native procedures); complete SQL hand rewrite with local envelopes 7.26 ms. The implemented generic rewrite now reaches that range automatically. Current cost coefficients select the faster existing operator once equality and local predicates are visible; no coefficient adjustment was necessary.

Local raw artifacts: `tpch/runtime/q19-final-{baseline,candidate}/perf.json`, `tpch/runtime/outliers-q19-guards/`, `tpch/runtime/integrated-20260915/`, and `tpch/runtime/q19-q13-analysis/` (untracked developer artifacts).

### Validation

- Boolean regressions: 13/13 with JIT, including all 27 TRUE/FALSE/NULL input combinations, unrestricted OR branches, duplicate join matches, outer-join NULL extension and invalid-column binding.
- TPC-H analytic correctness suite and correlated LEFT JOIN suite pass with JIT.
- Both new performance cases validate exact fixture revenue 5195.
- Scheme formatter leaves the modified file unchanged. Existing line-247 warning also occurs on baseline; no new warnings. `git diff --check` passes.
- Full tests and performance A/B are delegated to CI as requested. #908's latest JIT CI has a separate `found pointer to free object` process crash during Q13; targeted local TPC-H rerun passes, so that stability issue remains open and is not claimed fixed here.


carli2:

PR #912, opened 2026-09-15
Reduce grouped-query copying with bounded accumulator ownership

carli2:
Grouped outer joins such as TPC-H Q13 repeatedly copy growing dictionaries when the compiler discards accumulator ownership for large fused callbacks. Immutable `merge_assoc` also copied the whole destination for every source key, and group insertion batches repeatedly copied their accumulated rows.

This change proves a coarse ownership invariant once for large reducers, retaining conservative behavior for borrowed returns and nested speculative analyses. Immutable merges copy their destination once; group insertion uses a flat append accumulator. Regression tests cover input preservation, exceptions, borrowed returns, bounded analysis, duplicate customer keys, NULLs, and equivalent preaggregation shapes.

The branch includes the #908/#910/#911 stack and the prerequisite computed-column fix that materializes callback values once for compression analysis and encoding. It targets current `master` (68f8092b6); CI therefore checks the combined changes. This incremental optimization adds no planner dominance rule.

### Manual development A/B

Development baseline: 9f4401546, including the preceding correctness/performance stack. Candidate: 99ef1428c. Same AMD Ryzen 9 7900X3D, JIT enabled, GOMAXPROCS=24, isolated rebuilt Memory fixtures, 2 warmups and 5 timed samples, median. Fixture workload hashes match between baseline and candidate.

| Fixture (512 customers plus duplicate/NULL cases, 8192 orders) | Baseline | Candidate | Change |
|---|---:|---:|---:|
| Original outer count grouping | 93.660 ms | 74.545 ms | -20.4% |
| RHS preaggregated, duplicate-preserving outer grouping | 40.068 ms | 30.407 ms | -24.1% |
| Both inputs preaggregated with multiplicity | 33.004 ms | 31.417 ms | -4.8% |

The isolated Go immutable-merge benchmark (`-tags=nophp`, `-benchtime=1x -count=3`, median) at 4096 keys improves from 434.762 ms to 0.277 ms and from approximately 1.14 GB to 279 KB allocated per merge. These are operator measurements, not query speedups.

### Validation

- 212 targeted SQL cases pass with JIT: new fixture, group-stage corners, group-cache invalidation, TPC-H correctness shapes, and scalar-probe compile scaling.
- Targeted Go ownership/association/helper/reducer tests pass. The ownership regression was reproduced against the original compiler.
- The 48-scalar-probe compile gate completes in 87.7 ms with the final compiler.
- Full correctness and performance A/B remain in CI, as requested; no full local suite was run.

[...]

PR #913, opened 2026-09-15
Choose grouped join carriers and drivers by cost

carli2:
TPC-H Q18 repeatedly probes a grouped relation from the wrong join driver and rebuilds a query-backed group on every request. The planner now exposes the existing maintained group cache as a costed alternative, observes the required group's filtered cardinality once per request when repeated probes justify it, and lets logical join reorder choose the driver from that scalar observation. Warm Q18 falls from 11.12 seconds to 1.57 milliseconds without adding a storage operator.

This PR targets `master` (verified `68f8092b6`) and includes the prerequisite stack through #912 (`99ef1428c`); the new follow-up is the final commit. It can be reviewed against that baseline. The inherited stack includes TPC-H correctness fixes, join costing/predicate exposure, and group merge/ownership/compression improvements. Its outstanding global performance regressions are not claimed resolved here.

Implementation:
- Choose query-local grouping versus a maintained base-table group cache before assigning canonical physical cache identities; preserve logical stage IDs and consistent consumers.
- Feed only the observed filtered row count into logical reorder and its guards. Keep the prepared table in the existing request-scoped physical observation store, reused across guarded variants. Preserve the observation when the learned plan has moved the group to the first leaf.
- Price observation work against repeated group probes; a group already used once as the driver needs no observation. The required group build is not speculative work subject to the unrelated 100 ms speculation budget.
- Enable guarded plans for IN and parenthesized SELECT shapes, including equivalent derived-table comma joins. Prune guard reads for observations whose physical producer was not selected.
- Expose both group-input alternatives to EXPLAIN PHYSICAL CALIBRATE and Costgen's existing coefficient vector and ordering validation. Coefficients are unchanged; no hard dominance rule or global coefficient refit is added.

Manual A/B, performed before opening this PR:
Development baseline `99ef1428c` versus the proposed Scheme/compiler changes; same JIT binary, existing SF0.1 `.tbl` files, fresh isolated instances, GOMAXPROCS=24, 2 warmups and median of 5 samples. HTTP PostgreSQL-compatible SQL endpoint, complete JSON result consumption, all 42 baseline/candidate responses checked against the existing oracle. No data regeneration and no `/tmp` workspace.

| Query | Baseline ms | Candidate ms | Change |
|---|---:|---:|---:|
| Q18 | 11122.003 | 1.568 | -99.99% |
| Q13 | 1697.374 | 1636.676 | -3.58% |
| Q19 | 6.417 | 6.853 | +6.80% |

First Q18 request (fresh server, including compilation/cache construction): **11247.946 → 548.892 ms**. The warm result relies on the maintained group cache; it is not a claim about uncached one-shot latency.

The hand-written alternatives were measured through the same SQL endpoint before implementation: group-first query-backed input **740.057 ms**, group-first reusable base input **1.228 ms**, old order with reusable base input **5418.677 ms**. All hand-plan results matched the oracle. EXPLAIN, IR, PHYSICAL and REORDER artifacts were captured.

The new fixed 8192-line-item regression fixture, identical setup/warmup/sample configuration on both revisions:

| Fixture | Baseline ms | Candidate ms | Change |
|---|---:|---:|---:|
| Perf: selective grouped membership with outer aggregate | 45.468 | 1.134 | -97.51% |
| Perf: equivalent comma join uses the same group alternatives | 43.634 | 1.482 | -96.60% |
| Perf: tiny grouped input retains cheap query builder | 0.828 | 0.924 | +11.63% |

The small-case overhead is retained in the report rather than hidden by the large gain. Costgen's forced variants hold their reordered input fixed; their whole-query measurements are not interchangeable with adaptive SQL execution, and these validation-only rows are not used to fit new group-build coefficients.

Validation:
- 246 checks passed across eight related SQL suites (groups/HAVING, invalidation, canonical query groups, cache growth, session isolation and group-stage corners).
- Updated focused suites pass, including exact observed count, operator choices, equivalent SQL spelling, empty HAVING, NULL keys/quantities, DELETE/INSERT/UPDATE invalidation, and must-fail invalid columns.
- `go test ./tools/costgen` passes; JIT Costgen validates all four forced alternatives for the large and tiny fixtures with equal results and matching physical operators.
- Go formatting, Scheme formatter and `git diff --check` run. Existing unrelated formatter changes were excluded.
- Full tests are deliberately delegated to CI, as requested. #912's full/JIT correctness suites passed, but both inherited global performance A/B jobs failed; this PR still needs its own full CI results and later stack-wide balancing.


carli2:

PR #914, opened 2026-09-15
Cost Q2 order delivery and repair join cardinality estimates

carli2:
Q2 previously forced a supplier-led ORDER BY prefix even when the physical plan subsequently sorted. Worse, the Scheme reader splits `1e300` into separate tokens, silently capping join fanout at one row. This change parses that ceiling explicitly (including the zero-cost IKKBZ sentinel) and compares the ordered-prefix winner with an unrestricted join plus materialization/sort costs.

The automatic SF0.1 plan now scans the singleton region and selective parts before keyed supplier lookups. LEFT JOIN semantics remain intact. EXPLAIN REORDER reports both costs and the order-delivery choice. Sort and projected-column costs reuse existing Costgen coefficients; no new physical operator or dominance rule is introduced. The additional search uses the existing statistics/parameter guards.

Manual A/B, measured before creating this PR:

| Workload | Baseline | Proposed | Change |
|---|---:|---:|---:|
| TPC-H Q2, SF0.1 | 1089.023 ms | 40.779 ms | -1048.244 ms (-96.26%), 26.7x faster |
| CI correlated-MIN fixture, 8,192 partsupp rows | 30.464 ms | 33.944 ms | +3.479 ms (+11.42%) |

Baseline: `2ae6159cc`; proposed: `60d7414e0`. Same host, JIT enabled and verified, GOMAXPROCS=24, two warmups and median of five samples. Each side used a fresh instance, identical setup/rebuild, query and EXPLAIN sequence. SF0.1 reused existing prepared TPC-H files. HTTP psql requests include response parsing; the small YAML fixture uses the normal HTTP SQL runner. All 14 SF0.1 responses match the 44-row oracle. The smaller fixture is a measured regression below the 20% gate, not a claimed win.

The development sequence included EXPLAIN/IR/PHYSICAL/REORDER, perf sampling and closed hand plans. Original automatic Q2 was 1,210.803 ms; part-first LEFT JOIN hand plans took 42.318–45.282 ms, while region-first then supplier took 104.803 ms. Fixing order delivery alone reached 107.952 ms; correcting the cardinality ceiling reached the hand-plan range. The extracted `scan_join_order` MIN producer measures 11.235 ms, emitting 11,985 grouped entries (23,970 flat association elements). A separate part-filter scan measured 2.212 ms; these isolated timings are diagnostic and are not additive end-to-end costs.

The previous same-machine PostgreSQL measurement is 13.417 ms, **not remeasured for this PR**: MemCP Q2 remains approximately 3.04x slower. Group preparation/materialization and residual join work remain follow-up opportunities. Cardinality estimates are still imperfect: the final estimate is about 25,930 rows versus 44 actual rows, so this does not claim complete cost-model calibration.

Targeted validation: correlated-MIN ties, NULL/empty domains and mutations; must-fail missing correlated column; cardinality bounds and runtime-expression agreement; both sides of order-delivery cost selection; nullable-stage reorder; ordered scans/lookups; cached-plan growth; scalar-probe compile scaling. These passed with JIT. Scheme formatting introduces no new warnings (11 pre-existing baseline warnings). Full tests and global performance A/B are delegated to CI as requested.

Targets `master` at `68f8092b6` and includes the #913 stack. The predecessor has failing full/JIT and performance CI checks; this PR is not a claim that the inherited stack is green. Consolidation and global cost balancing remain necessary before merge.


carli2:

PR #947, opened 2026-09-18
Fix duplicate correlated subquery stage rebuilds and a bind-placeholder identity gap

carli2:
## Summary
- Two structurally identical correlated subqueries (one referenced directly, one nested inside a sibling's own expression tree) previously rebuilt their materialized group-cache table once per occurrence instead of sharing one stage, causing a multiplicative plan-size blow-up on deeply nested dashboard-style queries — confirmed 3x+ on a minimal repro matching the shape of a live production slowdown.
- Root cause #1: `binding_scope_entries` renamed a subquery-expression's own FROM alias to a position-dependent name whenever an *ancestor* subquery happened to expose an alias of the same name — even though a subquery expression (EXISTS/IN/scalar) never flattens into any ancestor's FROM list, so that alias can never actually collide. Fixed by threading a new `outer_collision_relevant` flag through `bind_query_names_at`/`bind_query_block_names`/`binding_scope_entries`, false only across a subquery-expression boundary.
- Root cause #2: positional bind-placeholder naming (`v1`, `v2`, ... assigned by pure textual occurrence order) gave two placeholders holding the identical bound value different symbolic names, independently defeating stage dedup. Fixed via `canonicalize_session_placeholders`, hooked into `decorrelate_logical_query` right where the raw parsed AST enters decorrelation — deliberately not inside `lib/sql-parser.scm`'s `parse_sql`, since adding new local definitions there was empirically found to corrupt results of unrelated queries containing no placeholder at all (an optimizer-sensitive local-variable-numbering fragility in that specific function).

## Test plan
- [x] New regression test `tests/planner/subqueries/nested-scalar-alias-reuse.yaml`: correctness (alias-reuse doesn't corrupt correlation resolution) + a `max_plan_size` dedup guard (fails at 214853 > 190000 on an unfixed build, passes at ~165KB on the fix).
- [x] Full sweep of `tests/planner/{subqueries,aggregates,joins,set-operations}`, `tests/sql/expressions/prepared-stmts.yaml`, `tests/integration/regressions` against current `origin/master`: clean except two pre-existing failures confirmed to reproduce identically on an unmodified master build (`tests/planner/joins/semijoin-carriers.yaml`, `tests/planner/subqueries/deep-correlation-membership.yaml`'s three noncritical cases).


carli2:
Closing in favor of two narrower, independent replacements:
- #954 fixes stage-dedup identity for correlated subqueries generally, via structural (schema/table/position) canonicalization instead of alias-preservation — this supersedes this PR's `outer_collision_relevant` alias-collision-avoidance mechanism (stage identity no longer depends on the SQL-surface alias at all, so avoiding needless renaming is no longer load-bearing for dedup).
- #955 extracts and lands this PR's independent, unrelated bind-placeholder-by-value canonicalization (`canonicalize_session_placeholders`) on its own.

carli2:
Follow-up: #955 (the bind-placeholder canonicalization extracted from this PR) has been closed — a reviewer found it folds placeholder identity by runtime value in a way that can poison the value-independent SQL plan cache (`sql_queryplan_cache_key` in `lib/sql.scm`) with a value-specialized plan, causing silent wrong results on later executions with different bound values. That mechanism was carried over unchanged from this PR, so it has the identical flaw — please don't reopen or reuse `canonicalize_session_placeholder_key`/`canonicalize_session_placeholders` from this PR without first fixing that.

PR #952, opened 2026-09-19
Fix correlated subqueries with residual (non-equality) outer correlation

carli2:
## Summary

Fixes correlated subqueries whose only correlation to the outer row is a
residual (non-equality) condition -- e.g. `WHERE t.ts >= outer.bound`,
`WHERE t.ts BETWEEN outer.start AND outer.end` -- with no clean equality
lookup-pair key. Before this fix, such a subquery either crashed
(`Unknown function: nil`) or silently returned `NULL`/`0` instead of the
correct value, because the group-stage prepare pipeline mishandled these
correlations in several places:

- `stage_direct_prepare_semantic_candidate?` never checked for residual
  outer refs, so such a stage could be classified "eager" (prepared once,
  before any outer row is bound) instead of "lazy" (prepared per outer row).
- `residual_outer_refs` was folded into only one of six logical
  stage-builder functions; the other five never included it in their own
  keys/domain/lookup_keys.
- `group_key_equality_terms` built a nonsensical self-referential comparison
  for a residual key, which isn't a real local column of the stage's own
  table.
- `replace_group_session_expr` only substituted session-read occurrences
  inside compute-proxy column-init lambdas, not residual refs.
- `lazy_stage_prepare_binding` produced a malformed `(!begin)` with zero
  operands for a stage with no direct dependencies.
- `stage_prepare_key` (and `shared_prepare_owner_key`) used a purely static
  session key, so `once` froze the first outer row's population forever
  instead of gating per distinct residual value -- and even with a dynamic
  key, the `(session KEY (once (lambda () BODY)))` binding itself was
  constructed once in the query's shared top-level prelude, outside any
  per-row scope, so `KEY`/`BODY`'s direct references to the residual outer
  column were simply unbound there. Redesigned the lazy prepare path to use
  a genuine named function parameterized on the residual columns, with
  `get_or_compute_scoped` doing the per-distinct-value memoization -- the
  same pattern already used by `scalar_query_probe_recipe_binding`.
- `group_stage_domain_key_pairs` required every entry of a stage's
  `gs_domain` to resolve to one of that stage's own group keys, which is
  too strict once a value legitimately belongs to an outer query block
  instead (e.g. an outer `GROUP BY` key carried into domain for unrelated
  probe-binding reasons).

## Test plan
- [x] Verified against a realistic KPI-dashboard query shape: multiple
      independent `COALESCE(SUM/COUNT ...)` probes, each `BETWEEN`-correlated
      to a period table and equality-correlated to a cross-joined params row,
      return correct results across all periods.
- [x] `tests/planner/subqueries/prejoin-scalar-subselect.yaml`,
      `tests/planner/subqueries/subselect-in-clauses.yaml`,
      `tests/planner/aggregates/scalar-subselect-groupby.yaml` all pass.
- [ ] CI (full suite)


carli2:
Closing after rebase and architectural review. The branch has no committed regression for its motivating failure, still does not complete a minimal residual-only COUNT/BETWEEN correlation, and deterministically removes the existing costed scalar-group projection in nested-aggregate-carrier.yaml. Repairing this safely requires a smaller change built from a concrete failing regression; weakening the existing performance invariant would hide a real planner regression.

PR #954, opened 2026-09-19
Canonicalize correlated-subquery stage identity by structure, not alias

carli2:
## Summary
- `stage_id` (`stable_structural_hash` over a stage builder's subquery/keys/outer_domain/condition/aggregates tuple) previously hashed the raw, bound AST directly — including every SQL-surface alias spelling, so two structurally identical correlated subqueries with different aliases (the shape a query builder that assigns fresh per-occurrence aliases actually produces) never shared a stage or its materialized group-cache table.
- Reuses the same idea `group_table_name`/`canonical_helper_sources` already apply to the physical group-cache table name (identity by schema/table/position, never by disposable alias), one layer earlier at logical stage-id time, via new `canonical_signature`/`canonical_alias_map`/`canonical_signature_expr` helpers in `lib/queryplan-logical.scm`.
- Applied at every correlated-subquery-family stage-id site: scalar-agg, scalar-once, scalar-single, scalar-group-top, exists, exists-group, in, in-candidate.
- Added `tests/planner/subqueries/canonical-stage-hash-alias-diversity.yaml`, covering alias *diversity* (two occurrences, distinct aliases, identical structure).

## Relationship to other open PRs
- **#947** (`fix/nested-correlated-stage-dedup`) is closed in favor of this PR (its alias-collision-avoidance mechanism is now redundant) plus #955... **update: #955 was closed** — the bind-placeholder-by-value part of #947 had a serious plan-cache-poisoning bug (folds by runtime value, which can leak a value-specialized plan into the value-independent SQL plan cache). Not ported.
- **#938** and **#952** are unaffected by this change (different concerns: plan-cache guard compilation, and residual/inequality correlation correctness respectively). This PR targets `master` directly and does not depend on either.

## Real-world measurement
Measured against a real customer KPI dashboard query (~15-20 independently correlated metrics): the alias-canonicalization alone does **not** fix its ~35-minute runtime — `EXPLAIN IR` stage count barely moves (258 → 252 unique stages). That query's metrics are independently correlated to different parent rows, not aliased copies of one shape. The follow-up `ags`-exclusion commit (below) is a second, more direct lever for that query's shape (shared SUM/COUNT-style metrics over identical conditions).

## Follow-up commit (2nd commit on this branch)
Fixes a CI failure found on the first commit and adds a second, independent improvement:

1. **CI fix**: `normalize_stage_dependencies_stages`/`_union_block` (`queryplan-optimize.scm`) flattened nested stages using `merge_unique` (whole-value equality) instead of `unique_stages_by_id` (gs_id-key equality) — harmless before this branch (raw-alias hashing meant two "duplicate" stages never coincidentally shared a gs_id), but exposed once canonicalization correctly converges structurally-identical subqueries. Fixed by switching both to `unique_stages_by_id`.
2. **`ags`-independent stage identity** for scalar-agg/scalar-single: two metrics differing only in aggregate function (e.g. COUNT vs SUM) over the exact same correlated condition now share one materialized carrier instead of each rebuilding its own. `merge_same_id_stage_variants`'s guard was relaxed from a one-directional superset check to a plain union (`same_group_carrier_backbone?` — full input/domain/keys/having/order/limit/offset equality — is the real safety net), and now recomputes `preserve_empty_domain` from the merged aggregate list.
   - `scalar-once` deliberately **keeps** `ags` in its hash: ORDER BY direction/OFFSET are baked into each descriptor and determine *which row* gets picked per key, not just which columns get read off an already-agreed selection. Attempting the same exclusion there surfaced a **pre-existing, independent bug already on unmodified master**: `stable_structural_hash` doesn't distinguish two scalar-once subqueries whose only difference is ORDER BY direction (ASC vs DESC) — confirmed to reproduce on pristine master, silently returning the same picked row for both. Left scalar-once as-is (matching master's current, already-buggy behavior) rather than fix that under this change; flagging it separately as its own follow-up.

One test in `traced-union-scalar-probes.yaml` (`Nested session-only presence probe is bound once per query`) is still failing — root cause not yet found, isolated by testing to be unrelated to either fix above.

## Test plan
- [x] New `tests/planner/subqueries/canonical-stage-hash-alias-diversity.yaml` (3 cases: alias diversity, SUM+COUNT merge, different-HAVING-stays-distinct) passes
- [x] `acl-boolean-membership-scaling.yaml` (previously CI-red) now passes
- [ ] `traced-union-scalar-probes.yaml` — 1 of 13 still failing, not yet root-caused
- [ ] CI full suite


carli2:
Closing this approach after review: it combines stage-identity canonicalization with aggregate-carrier merging, still has deterministic planner regressions, and did not materially improve the motivating dashboard workload. The semantic cache-identity change is too broad to merge safely in this form.

PR #955, opened 2026-09-19
Canonicalize bind-placeholder identity by value for stage dedup

carli2:
## Summary
A bind-placeholder name ("v1", "v2", ...) is assigned purely by textual occurrence order (`lib/sql-parser.scm`'s "?" rule and MySQL's own `COM_STMT_EXECUTE` bind-variable naming), with no regard for value. Two placeholders that happen to carry the identical bound value at compile time are, per Neumann's dependent-join domain D (a duplicate-eliminating set of correlation VALUES, not of positions), one and the same correlation value — but plan-level stage dedup (`unique_stages_by_id`/`stable_structural_hash`) compares the unevaluated expression tree, so two distinct names for an equal value silently defeat it.

Canonicalizes right where the raw parsed AST enters decorrelation, once per compile, by first-seen value (`canonicalize_session_placeholders` in `lib/queryplan-logical.scm`, hooked into `decorrelate_logical_query`). Deliberately kept out of `lib/sql-parser.scm`'s `parse_sql`: that function's optimizer-sensitive local-variable numbering breaks in surprising, unrelated ways when new local definitions are added to it, so the fix lives in a separate, small function instead.

## Relationship to other PRs
Extracted from #947, which bundled this fix with an alias-preservation fix (`outer_collision_relevant`) now superseded by #954's structural stage-hash canonicalization — alias identity no longer matters for stage dedup at all once #954 lands, so #947's narrower alias-collision-avoidance mechanism became redundant. This part (bind-placeholder canonicalization) is independent and unrelated to alias identity, so it lands on its own. #947 is being closed in favor of #954 + this PR together.

## Test plan
- [x] `tests/sql/expressions/prepared-stmts.yaml` (57/57)
- [ ] CI full suite


carli2:
Closing per reviewer finding, confirmed against the actual plan cache mechanism: `sql_queryplan_cache_key(username, schema, query, shape_hash)` in `lib/sql.scm` caches the compiled plan **without** the current session's bound placeholder values in the key — the entire point of a prepared-plan cache is to compile once structurally and reuse it across different bound values. `canonicalize_session_placeholders` folds placeholder identity by the *runtime value at compile time*, so if v1==v2 happens to hold on whichever session first triggers compilation of a given query shape, the resulting plan gets permanently specialized for "v1 and v2 are the same value" and stored under a value-independent cache key — every later execution with v1≠v2 then silently gets the wrong, mis-specialized plan.

This was ported faithfully from #947 without re-vetting its cache-safety, which was a miss on my part. Not attempting a quick patch here — folding placeholder identity by value must never be allowed to influence anything that ends up behind a value-independent cache key, which likely needs a different design (e.g. scoping the fold to be provably cache-key-transparent, or not folding by value at all for anything that reaches the plan cache).

